PR #167, opened 2021-12-05
Make sessions more persistent with re-authentication prompt

### LINKED ISSUE #166: [Feature] More persistent sessions
**Describe the feature**

Sessions are too fickle.  Storing everything in memory actually prevents users from navigating directly to another user's profile while signed in.  Signing in once the user has reached the desired profile takes the user away from that profile page, so the user is kind of stuck.

Store user sessions in browser session storage, but keep the user's plaintext keys stored in-memory.  Perform a re-authentication if the user has a session, but is missing their plaintext keys.

Jack-Edwards:
Resolve #166 

When I removed the "beta key" requirement from Crypter, I also chose to begin storing the user's secrets (JWT, private keys) in memory rather than use session storage or local storage.  This was for security reasons - I did not want to put any new users at risk by storing their sensitive information, in plaintext, in a location where malicious javascript has access to.

This pull request adds the concept of "re-authentication" and stores **non-sensitive** information in session storage and local storage again.

Items stored in session storage or local storage are:

- User ID
- Username
- Encrypted JWT
- Encrypted private keys

Items stored in memory are:

- Plaintext JWT
- Plaintext private keys

The "Remember me on this device" switch, now located on the login page, controls whether information will be stored in session storage **or** local storage.  This pull request does not implement refresh tokens.  That will be accomplished later.

Expected behavior:

Once a user is logged in, refreshing the page will prompt the user to re-authenticate.

If a user has chosen "Remember me on this device" when signing in, the user will also be prompted to re-authenticate when they open Crypter in a new tab.  If the user did not choose this option, the user will not be signed in on the new tab.  This is the difference between using session storage and local storage.

### DIFF
--- a/Crypter.API/Controllers/UserController.cs
+++ b/Crypter.API/Controllers/UserController.cs
@@ -139,8 +139,8 @@ namespace Crypter.API.Controllers
             {
                new Claim(ClaimTypes.NameIdentifier, user.Id.ToString())
             }),
-            Audience = "crypter.dev",
-            Issuer = "crypter.dev/api",
+            Audience = "www.crypter.dev",
+            Issuer = "www.crypter.dev/api",
             Expires = DateTime.UtcNow.AddHours(1),
             SigningCredentials = new SigningCredentials(new SymmetricSecurityKey(TokenSecretKey), SecurityAlgorithms.HmacSha256Signature)
          };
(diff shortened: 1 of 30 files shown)


Jack-Edwards (commented):
Not ready for acceptance.  Reworking a few things.

Jack-Edwards:
Still going to work on this.  I don't think re-authenticating should actually authenticate with the API.  The modal just needs to decrypt information stored locally.

Jack-Edwards:
Ready to review!

Issue #187, opened 2021-12-18
[Feature] Some columns should be made indexes in the database


Jack-Edwards:
**Describe the feature**

To make database queries more efficient, some tables could benefit from having some of their columns be made into an index.

- FileTransfer.Sender
- FileTransfer.Recipient
- MessageTransfer.Sender
- MessageTransfer.Recipient
- UserX25519KeyPair.Owner
- UserEd25519KeyPair.Owner
- UserToken.Owner

The first thing to do is modify the existing "Create_Foo.sql" scripts in the Crypter.Console project.  These table creation scripts are run whenever someone wants to scaffold a new database.  There should be lines in these scripts to add the indexes.

The second thing to do is create a new migration.  To do this, find the existing "MigrateX.sql" scripts in the Crypter.Console project.  Using the same patterns from an existing migration, create a new migration to add these indexes, if they do not exist.

PR #188, opened 2021-12-18
Added database index for better query efficiency on various tables

NNowakowski:
As per #187 I've added various DB indexes to the mentioned tables to increase efficiency on queries.
I couldn't find UserToken.Owner though; I assumed this would be UserProfile.Owner but didn't touch that one yet as I was unsure about that assumption.


Jack-Edwards:
Hey Nico, thanks for the contribution!

The UserToken table is something I will be merging in to the project soon.  I was not expecting anybody to pick up #187 before I merged the new table in.  I'll take care of adding the index to UserToken.

Jack-Edwards (approved):
Code looks good and passed testing on my end.  Nice work!

PR #191, opened 2021-12-21
Add new first touch page loader

### LINKED ISSUE #157: [Feature] Implement a better "Loading..." experience
**Describe the feature**

When a user first navigates to the Crypter website, they are greeted with a page that only contains a "Loading..." message.

See if this page can be customized.  If so, implement something a bit better.  A basic spinner would be an improvement.

jakubstankowski:
Resolve #157 

I implemented some new loading page, before user see the Crypter.dev page
This contains a name of app, description, spinner and loading info
Its looking like in the picture below

Zrzut_ekranu_456


Jack-Edwards (approved):
The html changes look solid.  I am still undecided the strings/literals we want to display while the application is loading, but that can be easily changed later.

Nice work!

Jack-Edwards:
Changed the target branch to "dev" instead of "main".

PR #195, opened 2021-12-23
Logout should update UI

Jack-Edwards:
At some point during development, I must have modified the AuthenticationService's "Logout" method to be asynchronous.  I forgot to await this method when the user clicks the Logout button.  This caused some problems.

Logout is now awaited.

Other changes:
* Added methods to the HttpService that only return an HTTP status.  These methods do not return any data.  These were required specifically for the Logout call to the API, which does not return any data.
* Fixed the page footer not sticking to the bottom of the page.


Jack-Edwards:
@jakubstankowski The website's sticky footer was no longer sticking to the bottom of the page.  This should be fixed in the commit named, "Fix footer not sticking to bottom of page".  Could you take a look to make sure I am not breaking the new loading UI?

PR #216, opened 2022-01-10
Implement customizable expiration times

### LINKED ISSUE #194: [Feature] Customizable expiration times
**Describe the feature**

Crypter automatically expires (and deletes) all items after 24 hours.  However, users may want to define their own expiration times.

Give users the ability to define when their uploads should expire.  24 hours can remain as the default expiration and the maximum expiration.

I personally do not care whether we allow users to set the expiration time down to the millisecond - the UI just needs to be clean and easy to use.

**Implementation Notes**

I would begin by modifying the Crypter.Contracts -> `ITransferRequest` interface.  Add a `RequestedExpiration` property to indicate when the user would *like* the transfer to be expired.  The API will need to verify the requested expiration is valid.

After that, add UI to Crypter.Web that allows users to define a custom expiration time.

Finally, Crypter.API needs to handle the new property.  The value should be validated and used in-place of the default 24 hours.

**Bonus**
Currently, Crypter.Console is run every 5 minutes using `cron`.  This is how we periodically delete expired items.

I would like to stop using `cron` and Crypter.Console for this.  I think Hangfire is a better option.

jakubstankowski:
Resolve #194 

I implement new customizable expiration times, inside the new  transfer settings section. I describe this in the video below

https://user-images.githubusercontent.com/36790120/148800270-54afad5a-a93f-4a43-91f7-715c2722683c.mp4

Let me know about implementation and UI design


Crypter.Web/Services/RequestedExpirationService.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
```
jakubstankowski, line 1: @Jack-Edwards  what do you think is good approach to use service for that method? I use this in two different places, and i dont want repeat the code
I also think about add some unit test for that method

Crypter.Web/wwwroot/css/app.css
```diff
@@ -51,4 +51,10 @@ a, .btn-link {
 /* BEGIN CUSTOM STYLES */
 .word-wrap {
     white-space: pre-line;
+}
+
+.upload-settings {
+    max-width: 600px;
+    margin: auto;
+    text-align: left;
```
Jack-Edwards, line 59: There should be some bootstrap classes we can use.

Crypter.Web/Services/RequestedExpirationService.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public DateTime ReturnRequestedExpirationFromRequestedExpirationInHours(int hours);
+    }
+
+
+    public class RequestedExpirationService : IRequestedExpirationService
+    {
+        public DateTime ReturnRequestedExpirationFromRequestedExpirationInHours(int hours)
+        {
+            return DateTime.UtcNow.AddHours(hours);
```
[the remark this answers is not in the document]
Jack-Edwards: I like reducing the amount of duplicate code in the project, but creating a service for this one method is a little overkill.  I also do not think the Settings component needs to inherit from the UploadTransferBase class.

Here is an example of a different approach that I might take: https://github.com/Crypter-File-Transfer/Crypter/commit/0e8a968bc9b5fbe9eda6187a59924cb70d95392b

Crypter.Web/Shared/Transfer/UploadMessageTransfer.razor
```diff
@@ -51,7 +53,7 @@
          <button type="button" class="btn btn-primary mx-auto d-block" @onclick="@(e => OnEncryptClicked())">Encrypt</button>
       </div>
    </div>
-   <TransferSuccessModal @ref="ModalForAnonymousRecipient" UploadType="message" ItemId="@TransferId" RecipientX25519PrivateKey="@EncodedRecipientX25519PrivateKey" ModalClosedCallback="@UploadCompletedEvent" />
+   <TransferSuccessModal @ref="ModalForAnonymousRecipient" UploadType="file" ItemId="@TransferId" RecipientX25519PrivateKey="@EncodedRecipientX25519PrivateKey" ModalClosedCallback="@UploadCompletedEvent" RequestedExpirationInHours="RequestedExpirationInHours"/>
```
Jack-Edwards, line 56: Make sure this gets changed back to "message"

Crypter.Web/Shared/Transfer/UploadTransferBase.cs
```diff
@@ -165,5 +176,24 @@ protected virtual void Cleanup()
...
+
+        protected Task OnRequestedExpirationChange(ChangeEventArgs e)
+        {
+            if (!string.IsNullOrEmpty((String)e.Value))
+            {
+                var value = Convert.ToInt32((String)e.Value);
+
+                if (value <= 0 || value > 24)
+                {
+                    requestedExpirationInHours = 24;
```
Jack-Edwards, line 188: I forgot to consider this scenario in the example code I linked in my other comment.

Crypter.API/Services/UploadService.cs
```diff
@@ -119,9 +119,8 @@ private async Task<(UploadResult Result, IBaseTransferItem GenericTransferData,
 
          Guid itemId = Guid.NewGuid();
          var itemCreated = DateTime.UtcNow;
-         var itemExpires = itemCreated.AddDays(1);
 
-         var returnItem = new BaseTransfer(itemId, senderId, recipientId, originalCiphertextBytes.Length, request.ClientEncryptionIVBase64, request.SignatureBase64, request.X25519PublicKeyBase64, request.Ed25519PublicKeyBase64, serverIV, serverDigest, itemCreated, itemExpires);
+         var returnItem = new BaseTransfer(itemId, senderId, recipientId, originalCiphertextBytes.Length, request.ClientEncryptionIVBase64, request.SignatureBase64, request.X25519PublicKeyBase64, request.Ed25519PublicKeyBase64, serverIV, serverDigest, itemCreated, request.RequestedExpiration);
```
Jack-Edwards, line 123: We should add a check at the very top of this method to verify the requested expiration DateTime is between 1-24 hours in the future.

Crypter.Web/Services/RequestedExpirationService.cs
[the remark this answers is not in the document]
Jack-Edwards: Pushed a new commit to my "examples" branch.  Take a look: https://github.com/Crypter-File-Transfer/Crypter/commit/b3ccee5fb8c2fee9973bb0628b60a21e1b41a7d1

The code to convert `int RequestedExpirationHours` to a `DateTime` can be handled by the `UploadTransferBase` component that both of our upload components inherit from.

Crypter.Web/Shared/Transfer/TabbedUploadTransferComponent.razor.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Microsoft.AspNetCore.Components;
```
Jack-Edwards, line 1: This file will need a copyright/license comment header.

Crypter.Web/Shared/Transfer/TransferSettings.razor.cs
```diff
@@ -0,0 +1,36 @@
...
+                _requestedExpirationHours = value;
+
+                RequestedExpirationHoursChanged.InvokeAsync(value);
+            }
+        }
+
+        [Parameter]
+        public EventCallback<int> RequestedExpirationHoursChanged { get; set; }
+
+        protected int requestedExpirationInHours = 24;
```
Jack-Edwards, line 29: Is this field being used?

Crypter.Web/Shared/Transfer/TransferSettings.razor.cs
```diff
@@ -0,0 +1,36 @@
+﻿using Microsoft.AspNetCore.Components;
```
Jack-Edwards, line 1: This file will also need a copyright/license comment header.

Crypter.Web/Shared/Transfer/UploadMessageTransfer.razor
Jack-Edwards: Looks like this still needs to be changed back to "message".

Jack-Edwards:
@jakubstankowski Sorry again for going straight to providing a solution, rather than giving you good feedback to make your own changes.  Thank you for being candid about that - I will be better in the future.

I marked a few of my previous comments as resolved.  Please take a look at the remaining comments and let me know what you think.

Crypter.API/Services/UploadService.cs
jakubstankowski: i add some basic  DateTime validation above this method

Crypter.Web/Shared/Transfer/UploadMessageTransfer.razor
jakubstankowski: its changed

Crypter.API/Services/UploadService.cs
Jack-Edwards: This check should be performed before we invoke `ItemDigestFunction()` or `SimpleEncryptionService.Encrypt()`.  This will prevent the server from performing the `ItemDigestFunction()` and `SimpleEncryptionService.Encrypt()` functions, which are heavy operations.

I would also prefer that rather than throwing an exception, return a new error code, similar to `return (UploadResult.InvalidServerEncryptionKey, null, null);` on line 115.  That is an existing pattern in the API.

If throwing an exception is a better pattern, let's implement that throughout the API at a later date.

Crypter.Web/Pages/UserSettings.razor
```diff
@@ -34,6 +34,11 @@
    @if (!Loading)
    {
       <h1>Settings</h1>
+       <div class="text-center">
+        <img src="http://ssl.gstatic.com/accounts/ui/avatar_2x.png" class="avatar img-circle img-thumbnail" alt="avatar">
+        <h6>Upload a different photo...</h6>
+        <input type="file">
+        </div>
```
Jack-Edwards, line 41: Can this be reverted?

jakubstankowski: its already reverted, it's should disappear after confirm pull request

Crypter.API/Services/UploadService.cs
jakubstankowski: i refactor this, i think better is throwing exception, because now API return strange data, with 400 error code like in code example below:
`{
    "result": 9,
    "id": "00000000-0000-0000-0000-000000000000",
    "expirationUTC": "0001-01-01T00:00:00"
}`

in my solution i return some clear error message, its easier in frontend whats going wrong

but let's refactor this in future

Jack-Edwards (approved):

PR #217, opened 2022-01-15
Additional unit tests for crypto code

Jack-Edwards:
Added unit tests for some methods within Crypter.CryptoLib that did not already have any tests.

In addition, I moved the one method contained in 'CommonCrypto.cs' to 'ECDH.cs'.  That method is only intended to be used in situations where a Diffie-Hellman key exchange has already been performed, so it seemed appropriate.


Crypter.CryptoLib/KeyConversion.cs
```diff
@@ -42,11 +42,6 @@ public static string ConvertToPEM(this AsymmetricKeyParameter keyParams)
          return stringWriter.ToString();
       }
 
-      public static byte[] ConvertToBytes(this KeyParameter symmetricKey)
```
Jack-Edwards, line 45: This method was not being used.

PR #221, opened 2022-01-24
Begin using MediatR

Jack-Edwards:
Begin using MediatR to handle negotiation between the API and the database.  I would like to slowly phase out the various "services" that exist in Crypter.Core with CQRS "commands" and "queries".

### DIFF
--- a/Crypter.API/Controllers/AuthenticationController.cs
+++ b/Crypter.API/Controllers/AuthenticationController.cs
@@ -28,9 +28,11 @@ using Crypter.API.Services;
 using Crypter.Contracts.Enum;
 using Crypter.Contracts.Requests;
 using Crypter.Contracts.Responses;
+using Crypter.Core.Features.User.Commands;
+using Crypter.Core.Features.User.Queries;
 using Crypter.Core.Interfaces;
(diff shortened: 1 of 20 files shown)


Crypter.API/Controllers/AuthenticationController.cs
```diff
@@ -81,32 +77,30 @@ public class AuthenticationController : ControllerBase
...
       public async Task<ActionResult<LoginResponse>> AuthenticateAsync([FromBody] LoginRequest request, CancellationToken cancellationToken)
       {
-         var user = await _userService.AuthenticateAsync(request.Username, request.Password, cancellationToken);
-         if (user == null)
+         LoginQueryResult loginResult = await _mediator.Send(new LoginQuery(request.Username, request.Password), cancellationToken);
+         if (!loginResult.Success)
          {
-            return new NotFoundObjectResult(
-               new LoginResponse(default, default, default));
+            return NotFound("Invalid username or password");
```
Jack-Edwards, line 83: @jakubstankowski As I replace calls to `_someService` with `_mediator.Send`, I will also modify the failure responses from the API.

PR #224, opened 2022-01-31
Refactor Client and API communication

Jack-Edwards:
This pull request standardizes the responses sent back by the API.  Every endpoint will response with it's own type of "success" response or a standard type of "error" response.  These responses are defined as classes in the Crypter.Contracts project.

"Success" responses are similar to the basic "response" classes we had before.  Some differences now are:

- A "success" response is only sent when the Http status is 200.
- "Success" responses no longer contain a "result" property, which could either indicate "success" or some type of error.
- Many "success" responses contain data to be used by the client.  Some "success" responses are empty classes.  Empty classes, rather than nothing at all, makes for cleaner infrastructure code.

"Error" responses:

- These are sent when the Http status is 400 or 404.  401 still responds with no content.
- To construct an "error" response, you must provide it with an enum value.  This sets an error code within the response and sets a human-readable error message.
- After deserializing an error response, the client must cast the error code back to the proper enum in order for the error code to mean anything useful.


Jack-Edwards:
This pull request is for preview purposes only.  I do not expect the the projects to build.

PR #225, opened 2022-02-08
Use MediatR in the Authentication controller

Jack-Edwards:
This pull requests replaces the calls to IUserTokenService in the Authentication controller with MediatR queries and commands.


Jack-Edwards:
Apparently I forgot some things.

PR #230, opened 2022-02-16
Try different patterns in the Authentication controller

Jack-Edwards:


Jack-Edwards:
@jakubstankowski Would you mind taking a look at this?  I am trying to make the `AuthenticateAsync` method as easy to understand as possible, while still making use of the `Either` class to return error / success data.  I am really curious what you think.

Issue #256, opened 2022-03-31
[Feature] Default "Remember me" to true


Jack-Edwards:
**Describe the feature**
Modify the login component so the "Remember me" option defaults to true.

For reference:
https://github.com/Crypter-File-Transfer/Crypter/blob/main/Crypter.Web/Shared/LoginComponent.razor.cs

ez-sherlock:
Please assign this to me. Thanks :)

Jack-Edwards:
@ez-sherlock Done.  Thanks for offering to take this on.  I updated the issue description with a link to the component.

ez-sherlock:
Can I get the issue's website link too?

Jack-Edwards:
Could you elaborate?  I don't think I understand which link you are asking for.

ez-sherlock:
I made the pull request. Please have a look at it. Thanks 💯

ez-sherlock:
@Jack-Edwards So sorry I just read the contribution guidelines. Should I delete this pull request and create a new one with the changes or is it fine by you?

Jack-Edwards:
Definitely don't merge this directly to the main branch. Modify this PR or submit a new one so the target branch is "dev".

Sent from ProtonMail mobile

-------- Original Message --------

ez-sherlock:
I again made a pull request. Have a look at it.

PR #258, opened 2022-03-31
Fixed - Input is always checked by default now

ez-sherlock:
FIxed the issue "[Feature] Default "Remember me" to true" 
Issue Link: https://github.com/Crypter-File-Transfer/Crypter/issues/256


Jack-Edwards (changes requested):
Unfortunately this fails to build.  The `@bind=` attribute is already responsible for controlling the `checked` state of the input.

What I would do is look at the code-behind (.razor.cs file) for this component and see if there is an appropriate place where we can set the `LoginModel.RememberMe` value to `true`.  The user should still be able to switch back to `false` if they choose.

ez-sherlock:
Okay let me make the pull request again :)

PR #262, opened 2022-04-02
"Remember Me" is now always checked by default at Login Page

ez-sherlock:
Link to the issue: https://github.com/Crypter-File-Transfer/Crypter/issues/256


Jack-Edwards (changes requested):
This also fails to build.  This modification needs to be made in the C# code, not the markup.

PR #363, opened 2022-07-06
Fixed bug 352 About page typo

Shawncologist:
This text: I had an Android phone so we I could not AirDrop it to her iPhone.


Jack-Edwards (approved):
Thanks for finding and taking care of this.  Good job!

Issue #508, opened 2023-04-25
Remove `_modalClassShow` from `ModalBehavior` front-end component


Jack-Edwards:
## Description

There is a variable in the `ModalBehavior.razor.z` code called `_modalClassShow`.  This variable has been around for years, but it doesn't appear to actually do or affect anything.

Verify the variable is unnecessary and remove it.

## Good First Issue

This is a good first issue for anyone interested in front-end development using ASP.NET and Blazor.

Exilon24:
I've removed **_modalClassShow** and I ran some of the unit tests and it doesn't seem to affect anything. It doesn't seem to be used anywhere either. The only problem is, I don't know how actually test the app to see if it works without that variable (I'm still new to this stuff 😅). Any advice?

Jack-Edwards:
Currently, none of the unit tests in the repo will verify the look or behavior of UI components.  But the front-end should be building successfully if you are able to run unit tests, so that is a good step.

Have you gone through this document on how to setup a local development environment?  [Development Environment Setup](https://github.com/Crypter-File-Transfer/Crypter/blob/main/Documentation/Development/Development%20Environment%20Setup.md)

For this issue specifically, it's sufficient to verify modal windows open/close and the background fades/un-fades on the front-end.  Every modal inherits from the class you are modifying, so it should be fine to test just one or two modals.  User registration and login should prompt a few modals to appear.

Jack-Edwards:
@Exilon24 If you want to give this another try, I suggest pulling the latest code from `main` or `stable` branch.  Run `docker-compose --profile dev up` in the solution directory.  The whole project should "just work" when you navigate to `https://localhost`.

PR #568, opened 2023-07-06
Removed _modalClassShow from ModalBehavior

aarondunn21:
Solved issue #508 by removing `_modalClassShow`, running unit tests, and verifying the front-end for no changes with docker

### DIFF
--- a/Crypter.Web/Shared/Modal/Template/ModalBehavior.razor
+++ b/Crypter.Web/Shared/Modal/Template/ModalBehavior.razor
@@ -26,7 +26,7 @@
 
 @inherits ModalBehaviorBase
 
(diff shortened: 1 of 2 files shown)


Jack-Edwards:
Thank you very much.  I'll verify this shortly.

aarondunn21:
> It looks like `ModalClass` can also be removed from the .razor and the .razor.cs files. Would you want to take care of that here a well?

Yes, I will take a look at those.

aarondunn21:
I have removed `ModalClass` and verified front-end functionality remains the same.

Jack-Edwards (approved):
Looks good.  Thank you.

PR #577, opened 2023-07-12
Rage contributions

rkj43:
@Jack-Edwards 
Here you go.
Also just to note, I have added the private key in Crypter.API/appsettings.json which I dont think is a good idea.

### DIFF
--- a/Crypter.API/appsettings.json
+++ b/Crypter.API/appsettings.json
@@ -44,14 +44,15 @@
          }
       ]
    },
-   "TokenSettings": {
-      "Audience": "www.crypter.dev",
-      "Issuer": "www.crypter.dev",
-      "SecretKey": "change me to something sufficiently long",
-      "AuthenticationTokenLifetimeMinutes": 5,
-      "SessionTokenLifetimeMinutes": 30,
-      "DeviceTokenLifetimeDays": 30
-   },
+  "TokenSettings": {
+    "Audience": "www.crypter.dev",
+    "Issuer": "www.crypter.dev",
+    "SecretKey": "change me to something sufficiently long",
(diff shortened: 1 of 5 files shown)


Jack-Edwards:
@rkj43 Did you try running the application?

rkj43:
> @rkj43 Did you try running the application?

Hi,
No I didnt.
I actually wanted to know if I was going in the right direction and then you asked me to submit a PR so you could check it, so I pushed without running it sorry.

Jack-Edwards:
Fair enough.  There is still some work to do to get the application in a running state, again.

I also noticed the old "symmetric" key is still in the Token Configuration.  That will need to be removed from the appsettings and the corresponding data model.

Once you remove the property from the data model, you should get some errors on build.  Those areas should be updated to consume either the EdDsa private or public key instead of the symmetric key.

rkj43:
> Fair enough. There is still some work to do to get the application in a running state, again.
> 
> I also noticed the old "symmetric" key is still in the Token Configuration. That will need to be removed from the appsettings and the corresponding data model.
> 
> Once you remove the property from the data model, you should get some errors on build. Those areas should be updated to consume either the EdDsa private or public key instead of the symmetric key.

Alright.
I will look into it.

rkj43:
Hey
Sorry for not getting back to you earlier.
 I was a bit busy with my thesis for my uni.
 
 I made the required changes but while testing several tests were failing and after some logging I found it was because of _SecurityTokenSignatureKeyNotFoundException_ . I am gonna give it another go and see if I can fix it.

Jack-Edwards:
Thank you for the update.  Ping me if you need any help.

PR #670, opened 2023-12-31
Enable 'Nullable' in several projects

Jack-Edwards:


Crypter.Common.Client/HttpClients/CrypterApiClient.cs
```diff
@@ -53,25 +50,34 @@ public class CrypterApiClient : ICrypterApiClient
...
     }
 
     public event EventHandler RefreshTokenRejectedEventHandler
     {
-        add => _refreshTokenRejectedHandler = (EventHandler)Delegate.Combine(_refreshTokenRejectedHandler, value);
-        remove => _refreshTokenRejectedHandler = (EventHandler)Delegate.Remove(_refreshTokenRejectedHandler, value);
+        add
+        {
+            _refreshTokenRejectedHandler = (EventHandler?)Delegate.Combine(_refreshTokenRejectedHandler, value);
+            UserAuthentication.RefreshTokenRejectedHandler = _refreshTokenRejectedHandler;
```
Jack-Edwards, line 75: Seems like this has been a bug for quite a while.  The instance of the event handler held by the UserAuthentication instance was not being kept up-to-date.

PR #703, opened 2024-04-19
Log an event to the EventLog table when a transfer is downloaded

Jack-Edwards:
Log an event to the EventLog table when a transfer is successfully downloaded.

Also fix an apparent bug; UserMessage transfers were not being deleted properly after download by the recipient.  `TransferItemType.File` was used instead of `TransferItemType.Message`.  Apparently this never impacted the live site, as the UserMessageTransfer table is currently empty.

### DIFF
--- a/Crypter.Core/Features/Transfer/Commands/GetUserMessageCiphertextCommand.cs
+++ b/Crypter.Core/Features/Transfer/Commands/GetUserMessageCiphertextCommand.cs
@@ -31,13 +31,14 @@ using System.Threading;
 using System.Threading.Tasks;
 using Crypter.Common.Contracts.Features.Transfer;
 using Crypter.Common.Enums;
+using Crypter.Core.Features.Transfer.Events;
 using Crypter.Core.MediatorMonads;
 using Crypter.Core.Repositories;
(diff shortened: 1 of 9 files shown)


Crypter.Core/Features/Transfer/Commands/GetUserMessageCiphertextCommand.cs
```diff
@@ -94,12 +92,14 @@ internal class GetUserMessageCiphertextCommandHandler
             return DownloadTransferCiphertextError.InvalidRecipientProof;
         }
 
-        bool deleteOnReadCompletion = !databaseData.RecipientId.HasValue;
-        return _transferRepository
-            .GetTransfer(id, TransferItemType.Message, TransferUserType.User, deleteOnReadCompletion)
-            .IfSome(_ => _backgroundJobClient
-                .Enqueue(() => _hangfireBackgroundService
-                    .DeleteTransferAsync(id, TransferItemType.File, TransferUserType.User, !deleteOnReadCompletion)))
```
Jack-Edwards, line 102: The wrong enum was present on this line.

PR #720, opened 2024-08-25
Multipart file transfers

Jack-Edwards:
Implement the capability to upload a large file in multiple parts, or across multiple requests.  This will supersede streamed uploads while streamed uploads are still broken in Chromium and unimplemented in all other browsers.


Jack-Edwards:
Back-end is complete.  Need to work on the HttpClient next.  Then finish the integration into the web client.

PR #752, opened 2025-01-11
Remove 'UploadNewKeys' from LoginResponse

Jack-Edwards:
The web client should stop depending on the LoginResponse to tell it to upload new keys.

Instead, the client should attempt to fetch the keys after login.  If the API responds 404, the client should create and upload new keys for that respective 404.

In other words, if GET MasterKey returns 404, then upload a new master key.  If GET PrivateKey returns 404, then upload a new key pair.

The intent is to decouple this key management from the login request/response.  Should make implementing MFA easier in a subsequent change.

### DIFF
--- a/Crypter.Common.Client/Events/RecoveryKeyCreatedEventArgs.cs
+++ b/Crypter.Common.Client/Events/RecoveryKeyCreatedEventArgs.cs
@@ -1,3 +1,29 @@
+/*
+ * Copyright (C) 2025 Crypter File Transfer
+ *
+ * This file is part of the Crypter file transfer project.
+ *
+ * Crypter is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * The Crypter source code is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * You can be released from the requirements of the aforementioned license
+ * by purchasing a commercial license. Buying such a license is mandatory
+ * as soon as you develop commercial activities involving the Crypter source
+ * code without disclosing the source code of your own applications.
+ *
+ * Contact the current copyright holder to discuss commercial license options.
+ */
+
 using Crypter.Common.Client.Models;
 
 namespace Crypter.Common.Client.Events;
--- a/Crypter.Common.Client/Events/RecoveryKeyCreatedEventArgs.cs
+++ b/Crypter.Common.Client/Events/RecoveryKeyCreatedEventArgs.cs
@@ -1,0 +1,13 @@
+using Crypter.Common.Client.Models;
+
+namespace Crypter.Common.Client.Events;
+
+public sealed class RecoveryKeyCreatedEventArgs
+{
+    public RecoveryKey RecoveryKey { get; }
+
+    public RecoveryKeyCreatedEventArgs(RecoveryKey recoveryKey)
+    {
+        RecoveryKey = recoveryKey;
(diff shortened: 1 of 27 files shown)


Crypter.Common.Client/Events/RecoveryKeyCreatedEventArgs.cs
```diff
@@ -0,0 +1,13 @@
+using Crypter.Common.Client.Models;
```
Jack-Edwards, line 1: Missing copyright header.

Crypter.Common.Client/Interfaces/Services/IUserKeysService.cs
```diff
@@ -36,10 +34,5 @@ public interface IUserKeysService
 {
     Maybe<byte[]> MasterKey { get; }
     Maybe<byte[]> PrivateKey { get; }
-
-    Task DownloadExistingKeysAsync(Username username, Password password, bool trustDevice);
-    Task DownloadExistingKeysAsync(byte[] credentialKey, bool trustDevice);
-    
-    Task<Maybe<RecoveryKey>> UploadNewKeysAsync(Username username, Password password, VersionedPassword versionedPassword, bool trustDevice);
-    Task<Maybe<RecoveryKey>> UploadNewKeysAsync(VersionedPassword versionedPassword, byte[] credentialKey, bool trustDevice);
```
Jack-Edwards, line 44: This is all handled internally by subscribing to the UserSessionService.UserLoggedInEventHandler event handler.

Crypter.Web/Shared/MainLayout.razor.cs
```diff
@@ -64,43 +64,30 @@ public class MainLayoutBase : LayoutComponentBase, IDisposable
...
+        if (args.ShowRecoveryKeyModal)
+        {
+            await RecoveryKeyModal.OpenAsync(args.VersionedPassword);
+        }
     }
 
-    private void ShowPasswordHashingModal(object? sender, PasswordHashBeginEventArgs args)
+    private void HandleRecoveryKeyCreatedEvent(object? _, RecoveryKeyCreatedEventArgs args)
+    {
+        RecoveryKeyModal.Open(args.RecoveryKey);
```
Jack-Edwards, line 87: This modal is appearing behind the "password hashing" model during first login.

Jack-Edwards: This is caused by `UserLoggedInEventArgs.ShowRecoveryKeyModal` also being set to `true`, therefore the MainLayout is instructing the RecoveryKeyModal to open twice.  The instance responding to the UserLoggedInEventArgs requires the modal to hash the user password, thus the password hashing modal appears.

Jack-Edwards: Same issue still occurs, except now the modal only says "An error has occurred".

Jack-Edwards: This issue is now resolved.

Although, there is now some delay between the PasswordHashModal closing and the RecoveryKeyModal showing.  Might need to introduce a new modal while the UserKeysService is handling the login event.

Jack-Edwards (commented):
crit: Microsoft.AspNetCore.Components.WebAssembly.Rendering.WebAssemblyRenderer[100]
      Unhandled exception rendering component: Missing sender private key
System.Exception: Missing sender private key
   at Crypter.Web.Shared.Transfer.UploadTransferBase.<>c.<SetHandlerUserInfo>b__57_1()
   at EasyMonads.Maybe`1[[System.Byte[], System.Private.CoreLib, Version=9.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].Match[Byte[]](Func`1 none, Func`2 some)
   at Crypter.Web.Shared.Transfer.UploadTransferBase.SetHandlerUserInfo(IUserUploadHandler handler)
   at Crypter.Web.Shared.Transfer.UploadMessageTransfer.OnEncryptClicked()
   at Crypter.Web.Shared.Transfer.UploadMessageTransfer.<BuildRenderTree>b__6_2(MouseEventArgs e)
   at Microsoft.AspNetCore.Components.ComponentBase.CallStateHasChangedOnAsyncCompletion(Task task)
   at Microsoft.AspNetCore.Components.RenderTree.Renderer.GetErrorHandledTask(Task , ComponentState )

Issue #572, opened 2023-07-08
[Feature] Add version information to the web client


Jack-Edwards:
## Description

I would like to know which version of the application I am using when I load www.crypter.dev.

Show the version number of the application somewhere on the front-end.

## Good First Issue

This may be a good first issue for someone interested in ASP.NET and Blazor applications.  This issue may also require some familiarity with Docker and/or Github actions.

## Additional Comments

Truthfully, I'm not sure of the best way to accomplish this task.  The release number isn't truly known until I publish a release in Github.  The release is tagged and Docker images are uploaded to Github Container Repository (GHCR).  You can see the Crypter.Web packages here https://github.com/Crypter-File-Transfer/Crypter/pkgs/container/crypter_web

Maybe we add a step in the release process that modifies the version number in the Crypter.web project's `.csproj` file?  As in, perform a simple string replace.

Showing the version number in the application should be a simple matter of showing the assembly version from the code.

Jack-Edwards:
I don't want to overload this issue too much, but keep in mind while working on the UI that we may also want to fetch and show the API version down the road.

Client version: 1.2.3
API version: 1.2.3

Ideally these numbers should always be the same, given the way our CI/CD is currently written.  But it would be extremely useful to know when these numbers do not match.  Did the API deploy fail?  Is someone using a cached version of the web client?

ibede:
Hi! As much as I saw from the current CI/CD, you may fetch the git refs for the current tag that can be then injected to the Dockerfile. In my opinion that can be then used during the build to set the Version of the dlls. After that it can be read in code. For the API project a new endpoint maybe for the version info is sufficient, for the client, it can be read locally. Also maybe caching the versions or a singleton "version service" would help.
If you think this approach is viable, I'll gladly implement it.

Jack-Edwards:
I'm on board with this approach.  On the API side, I'd still like to put the logic behind a Mediator handler for consistency, but the handler may be supported by a singleton service like you suggest.

What do you suggest when someone is running a non-release build?  Default to `0.0.0` or use the latest commit hash?

Lastly, it might be a nice touch to link the version shown on the front-end to the GH release page.  Not a requirement.

ibede:
Sure, I'll create a MediatR request handler for it.
For the non-release build I like the commit hash approach more, I think the AssemblyVersion contains it anyway by default - I'll check it whether it is available without specifying the version during build.
I think the link is feasible as well.

[...]

PR #761, opened 2025-02-07
Feature: Display version info (#572)

ibede:
* Injects versioning information into assemblies if available (tag, version, VCS URL, hash)
* Implements a singleton Version Service for the backend and new endpoint for fetching it
* Uses the same mechanism to get the frontend version info
* Displays version info in footer (popover)

Related to #572


Crypter.API/Dockerfile
```diff
@@ -20,8 +24,9 @@ COPY Crypter.Core/. ./Crypter.Core/
 COPY Crypter.Crypto.Common/. ./Crypter.Crypto.Common/
 COPY Crypter.Crypto.Providers.Default/. ./Crypter.Crypto.Providers.Default/
 COPY Crypter.DataAccess/. ./Crypter.DataAccess/
+COPY .git/. ./.git/
```
Jack-Edwards, line 27: Is this used?

ibede: Yes. Since .NET8, the git commit hash is automagically included ín the InformationalVersion during the build, if available. That's why I had to include it for build time.

Jack-Edwards: That makes sense.  Not sure that I want to include the git history in the Docker image just to support this feature though.  Is passing in another argument a possibility?

I can look into alternatives over the weekend.

Jack-Edwards:
Great work!  I'll have more time to review and run this over the weekend.  I appreciate your contribution.

Crypter.API/Dockerfile
ibede: Runtime does not contain it, it's only used during build ín the intermediate build container.

ibede: The other option would be to inject the hash as well via the CI action and bake it in the assembly into the new attribute, like the version control system URL.

Jack-Edwards: > Runtime does not contain it, it's only used during build in the intermediate build container.
I was just reading about this.

I prefer providing the hash via CI action for now.  Less to copy and a noticeably quicker build in the worst case scenario.

Crypter.API/Dockerfile
```diff
@@ -1,4 +1,8 @@
 FROM mcr.microsoft.com/dotnet/sdk:9.0 AS build
+ARG GIT_VERSION_TAG=0.0.0
+ARG GIT_VERSION_TAG_FULL=0.0.0
+ARG VCS_URL
+ENV VCS_URL=$VCS_URL
```
Jack-Edwards, line 5: I frequently build from docker compose: `docker compose up --build -d`.  Need to update the `docker-compose.yml` file?

Crypter.API/Dockerfile
ibede: Ok I'll fix it next week then. I'll leave the parsing part in for local native builds, there it works fine.

Crypter.API/Dockerfile
ibede: If you need version info in those builds, then you can pass those arguments to `docker-compose build` with `--build-args`. Also the args can be specified ín the `docker-compose.yml`

Jack-Edwards:
I've had a minute to sit down and think.

What do you think about using minver to set the file/assembly version with the latest tag during build?  [Link for example](https://github.com/adamralph/minver?tab=readme-ov-file#can-i-use-the-version-calculated-by-minver-for-other-purposes)

I think this would still require the `.git` folder to be copied during build, but that does seem like the most simple option.  I think we can forgive a few extra CPU cycles if it means less work to run the apps.

Crypter.API/Dockerfile
Jack-Edwards: Resolving this conversation in favor of https://github.com/Crypter-File-Transfer/Crypter/pull/761#issuecomment-2646079908

Crypter.Common/Infrastructure/VersionProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * Contact the current copyright holder to discuss commercial license options.
+ */
+
+using Crypter.Common.Attributes;
+using System.Reflection;
+
+namespace Crypter.Common.Infrastructure
+{
+    public record class ProductVersion(string Version, string? Hash);
```
Jack-Edwards, line 32: `class` is redundant here.

Crypter.Common/Infrastructure/VersionProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            Assembly asm = Assembly.GetEntryAssembly() ?? Assembly.GetExecutingAssembly();
+            VersionControlMetadataAttribute? metaData = asm.GetCustomAttribute<VersionControlMetadataAttribute>();
+            string? versionUrlBase = metaData?.BaseUrl;
+            
+            if (versionUrlBase != null)
+            {
+                string versionPath = isRelease ? RELEASE_PATH : COMMIT_PATH;
+                string version = isRelease ? productVersion : (hash ?? string.Empty);
+                return $"{versionUrlBase}/{(versionPath)}/{version}" ?? string.Empty;
```
Jack-Edwards, line 69: The left-hand string can never be null.  The right-hand `string.Empty` is unnecessary.

Crypter.Web/wwwroot/js/scripts.js
```diff
@@ -37,5 +37,19 @@
...
+    InitVersionPopover: function () {
+        const infoBtn = document.getElementById('info-icon');
+        if (infoBtn) {
+            const popoverOptions = {
+                content: document.getElementById('version-content'),
+                html: true,
+                placement: 'top',
+                trigger: 'click focus'
+            };
+            new bootstrap.Popover(infoBtn, popoverOptions);
```
Jack-Edwards, line 51: Neat!

ibede:
> What do you think about using minver to set the file/assembly version with the latest tag during build? [Link for example](https://github.com/adamralph/minver?tab=readme-ov-file#can-i-use-the-version-calculated-by-minver-for-other-purposes). If you want to check this out, make sure you install the NuGet in the assembly that you want to be versioned. I.e., the API and the web client; not the Common project.
> 
> I think this would still require the `.git` folder to be copied during build, but that does seem like the most simple option. I think we can forgive a few extra CPU cycles if it means less work to run the apps.

It's not a bad an option to use. It also supports prefix handling that is currently used for Crypter tags ('v'). Git info is needed though. What needs to be checked is whether the hash is added the same way as before or it needs to be manually added via `MinVerBuildMetadata`. What I don't see yet is the real advantage over the current solution. For docker builds the versioning system base URL has to be added anyway, maybe the hash too. For local builds it works the same way. Maybe it's more concise to handle the versions seamlessly within the build process without build properties.

Jack-Edwards:
A pre-release suffix sounds good to me.  Probably more accurate than the latest commit hash anyway, since the patch version will auto-incremented every time.  I'm in favor of using minver like you describe.

Sorry for the back-and-forth.  Thank you for helping me think through this.

ibede:
No problem. We should discuss this deeply, as the branching strategy also affects this, it's not a trivial decision.
In the meantime, I checked and the hash is still included in the InformationalVersion when using MinVer, so it's not necessary to add it manually. 
So at the end every build will then get a release version or pre-release version based on the git tag. Both locally and in docker. The pre-release version would be the last tag in the history, with incremented patch version and a pre-release suffix, like now it would be `1.21.1-pre.0.42`, where 42 is the commit height/distance from the last tag (1.21.0). For development it could be a bit misleading though as `stable` does not contain the previous tags in its history because of the branching strategy, but it would work fine for releases. To address this you may tag on `stable` or rebase `stable` after every release.
One thing I would change is the link for the commits - I think it would be better, if it opened the tree browser of that commit version in github, not the commit itself. So not [this](https://github.com/Crypter-File-Transfer/Crypter/commit/72e54497d841e49aba8d664027b6a7c6bbd1d784), but [this](https://github.com/Crypter-File-Transfer/Crypter/tree/72e54497d841e49aba8d664027b6a7c6bbd1d784)

Jack-Edwards:
I need to be rebasing stable.  It just hasn't mattered up to this point.

Jack-Edwards (approved):
Awesome work.  Thank you for doing this.

Issue #565, opened 2023-07-03
[Feature] Default the DbContext to use 'AsNoTracking'


Jack-Edwards:
I'd like to disable Entity Framework Core's tracking mechanism across the whole project; only enabling tracking in cases where the API would need to make an update an entity in the database.

ibede:
For this, I would suggest to explicitly set the `.AsNoTracking()` flag to those queries where it is relevant. By default EF does not track the entities if the query makes a projection to a runtime type if the entity itself is not included. I browsed through the feature `Queries` and most of the time Crypter returns a DTO or some anonymous object, not the entity itself. Those are not tracked. There are some places where it's not true and the whole entity is returned. Some `Commands` also do side-queries where the entities are not modified. I would only disable tracking at those places. Also, no-tracking when querying included entities can be tricky for identity resolution.
So I think disabling it generally and enabling for write situations is more error-prone, risky and also much more work.

Jack-Edwards:
I agree.  Probably slipped my mind at the time that projections would necessarily result in no-tracking behavior.  Closing.

ibede:
Yesterday, I prepared the changes for this.
https://github.com/ibede/Crypter/pull/1/

If you feel like, I can add them here still.

Jack-Edwards:
Looks good.  Go ahead and open a PR.

PR #766, opened 2025-02-13
Disable entity tracking where applicable

ibede:
* Use entity existence test instead of fetching the whole entity where applicable
* Use `AsNoTracking` on DbSets where applicable
* Remove unused DbContext references

Related to #565


Jack-Edwards (approved):
Great work.

Issue #575, opened 2023-07-08
[Feature] Switch JWT to public-key signature


Jack-Edwards:
## Description

If the plan is to implement a self-hosted storage option for Crypter, then the existing Crypter code needs to switch to signing JWTs with a private key, instead of a symmetric key.

This will let self-hosted applications verify the authenticity of JWT signatures using the corresponding public key that the Crypter API will need to make publicly available.

## Implementation notes

The code to sign and verify JWTs with a symmetric key is already written.  Really that just needs to be swapped out for public key.

~~An endpoint will also need to be added to the API, to expose the public key.~~ Edit: This won't be necessary until we implement OpenID Connect.

## Good First Issue

This may be a good first issue for someone who is interested in ASP.NET and is comfortable with JWT creation and verification.

rkj43:
Hey, I would like to try this out but I have never done any work in JWT creation and verification.
Is there any documentation or video or resource that might be helpful for the same?

Jack-Edwards:
With respect to the code, this is where the signing algorithm and key are configured:  https://github.com/Crypter-File-Transfer/Crypter/blob/main/Crypter.Core/Identity/TokenParametersProvider.cs

I would tackle this issue across multiple parts:

1. Read up on JWTs, understand why JWTs must be cryptographically signed, and understand the strengths and drawbacks of the different *asymmetric* signing algorithms.
2. Choose an asymmetric signing algorithm with support in .NET.  After consuming those articles below, it sounds like EdDSA is the best algorithm overall.  The down side is that ASP.NET does not yet support EdDSA out-of-the-box.  We would need to use Scott Brady's NuGet package.
3. Refactor the `TokenParametersProvider` class to use the new signing algorithm.  Update the backend's configuration with properties for a private key and public key (or calculate the public key during startup, from the private key).
4. Run unit tests.  There are a few tests to make sure tokens can be created and validated, but there are no tests that verify invalid tokens would be rejected.  Some additional test coverage here would be nice, granted we're touching this part of the code.

Down the road, I believe Crypter would also need to implement something like OpenID Connect in order to support self-hosted storage servers.  That's not going to be for a long time though - this doesn't need to be accomplished now as part of this issue.  Just thinking out loud.

### Good and relevant articles

A good intro to the different types of signing keys for JWT: https://www.scottbrady91.com/jose/jwts-which-signing-algorithm-should-i-use

[...]

PR #767, opened 2025-02-13
Feature: Asymmetric JWT signing

ibede:
* Implements EdDsa algorithm based JWT signing
* Exposes OpenID-compliant endpoints for external JWT signature validation purposes

Relates to #575


Crypter.Core/Identity/TokenKeyProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+ * Contact the current copyright holder to discuss commercial license options.
+ */
+using ScottBrady.IdentityModel.Crypto;
+using ScottBrady.IdentityModel.Tokens;
+
+namespace Crypter.Core.Identity
+{
+    public static class TokenKeyProvider
+    {
+        private static readonly EdDsa _algorithm = EdDsa.Create(ExtendedSecurityAlgorithms.Curves.Ed448);
```
Jack-Edwards, line 33: Thinking out loud, I would have probably ended up using Geralt's [digital signature](https://www.geralt.xyz/digital-signatures) feature.  Geralt is used elsewhere as the C# wrapper over libsodium.  But Ed448 does offer more bits of security, so probably a better choice.  I've referred to Scott Brady's blog posts quite a bit while working on this project.

Is it accurate to say this implementation still requires some work in terms of 1) persisting the private key and 2) deriving the public key from the private key, and not simply generating a random public key?

Crypter.API/Program.cs
```diff
@@ -57,29 +57,29 @@
         builder.Configuration
             .GetSection("AnalyticsSettings")
             .Get<AnalyticsSettings>()
-        ?? throw new Exception("AnalyticsSettings missing from configuration"),
+        ?? throw new ConfigurationException("AnalyticsSettings missing from configuration"),
```
Jack-Edwards, line 60: Thanks for the clean up.

Crypter.Core/Identity/TokenKeyProvider.cs
ibede: For the first question, I would say yes if you want to reuse the same key pair for every new deployment. The current implementation generates one at startup, so a next deployment enforces users to re-authenticate. If the same key should be used on next startup then the private key should be exported somewhere persistent, for example somewhere in the volume where the messages/files are stored. Also at startup it should be imported if available.
I'm not sure about the second one. The `EdDsa` already generates the key-pair upon construction, so the public and private keys are in sync. Maybe the misleading thing is that I wrap here the `_algorithm` into two separate keys. If my understanding is right, the `EdDsa` core is the same so both of them includes the same private and public keys, I just wanted to create two separate reference to it, for not to be mistaken. Here's the [implementation](https://github.com/scottbrady91/IdentityModel/blob/dc78d1988f2b36c5056acb2b2eeade37e08651cd/src/ScottBrady.IdentityModel/Tokens/EdDsa.cs#L64) in Scott's library. 
Scott's library wraps BouncyCastle.Crypto algorithms already, Geralt can be used of course, then we have to implement key handling by ourselves. By the way, creating a library for that based on Scott's lib could be a fine little spin-off project.

Jack-Edwards: Storing the persistent key in a volume makes sense.  I agree, the key should be imported at startup.

If a key is not found at startup, I'd say it's fine for development scenarios to proceed with a random key.  This behavior should be configurable for production scenarios.  Something like:

```
appsettings.json
{
    "JwtConfig": {
        "RequirePersistentSigningKey": true
        "PersistentSigningKey": // an env var
    }
}
```

```
appsettings.Development.json
{
    "JwtConfig": {
        "RequirePersistentSigningKey": false
    }
}
```

By the way, I'm getting ready to take my family camping this weekend.  Will be out of range for a while.

ibede: Yesterday I had some time to analyze deeper the key handling possibilities. It turned out that Scott's library does not implement importing and exporting the keypair. I also had a look on possibilities provided by Geralt. There, import/export is not in scope, so finally I think there are two possibilities here. If we want to support standard key formats like PKCS#8 with PEM encoding then we can use BouncyCastle's [PEMReader](https://github.com/bcgit/bc-csharp/blob/master/crypto/src/openssl/PEMReader.cs) and move on with Scott's library as it's only a wrapper around BC. Or we can switch to Geralt and implement own key encryption and persistence. To be honest I'm liking the Geralt more, but the disadvantage is that we cannot use keypairs generated in PEM formats (like ssh-keygen outputs).

ibede: I tried and finally managed to switch to Geralt's Ed25519 algorithm that is already used in the `Crypter.Crypto.Common.DigitalSignature`. Tomorrow I'll try to implement the key export/import

ibede:
So - at the end it became quite big.
New stuff: I implemented here also the persistence of the encrypted private key for the EdDsa. For this, I extended a bit the `IEncryption` interface and also implemented it for `DefaultCryptoProvider` with `ChaCha30Poly1305` from Geralt. The EdDsa algorithm and key handling has roots from Scott's library and also I had some inspiration from [Kryptor](https://github.com/samuel-lucas6/Kryptor) for the key persistence. Also added some tests.

Crypter.Core/Identity/TokenSettings.cs
```diff
@@ -24,14 +24,38 @@
...
+    Never,
+    WhenMissing,
+    Always
+}
 
 public class TokenSettings
 {
     public required string Audience { get; set; }
     public required string Issuer { get; set; }
-    public required string SecretKey { get; set; }
```
Jack-Edwards, line 33: This property should also be removed from the primary `appsettings.json` file.

Jack-Edwards: `Volumes/API/appsettings.json` as well, for local `docker compose` scenarios.

Crypter.API/appsettings.json
```diff
@@ -65,6 +65,10 @@
       "SecretKey": "change me to something sufficiently long",
       "AuthenticationTokenLifetimeMinutes": 5,
       "SessionTokenLifetimeMinutes": 30,
-      "DeviceTokenLifetimeDays": 30
+      "DeviceTokenLifetimeDays": 30,
+      "RequirePersistentSigningKey": true,
+      "PersistentSigningKeyLocation": "./Volumes/API/keys/jwt.sign.private",
+      "PersistentSigningKeyPassword": "Ch4ng3.Th15.V3rY.53cur3.Pa55w0rd!",
+      "SigningKeyGenerationStrategy": "WhenMissing" // Should be changed to "Never" after first startup.
```
Jack-Edwards, line 72: Need to add these settings to `Volumes/API/appsettings.json` as well.  Otherwise the API is failing to startup when running `docker compose --profile local up --build`.

Crypter.API/Controllers/JwtController.cs
```diff
@@ -0,0 +1,73 @@
...
+using Microsoft.AspNetCore.Http.Extensions;
+using Microsoft.AspNetCore.Mvc;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Crypter.API.Controllers
+{
+    [ApiController]
+    [Route(".well-known")]
```
Jack-Edwards, line 41: I think these paths should be pre-fixed with `api/`.

Not to say this setup is perfect, but I'm currently hosting both the web client and API from the same host and the same domain.  Caddy sends all requests for the `/api/` to the API container.  All other requests are given a copy of the web client, which takes over SPA routing.  Reference: https://github.com/Crypter-File-Transfer/Crypter/blob/main/Crypter.Web/Caddyfile#L9

Jack-Edwards: I think this is why `/swagger/index.html` is not working from Docker even when the environment is set to Development.

Crypter.Core/Identity/TokenSettings.cs
ibede: > `Volumes/API/appsettings.json` as well, which gets used in local `docker compose` scenarios.

This can only be done locally I think

Crypter.API/Controllers/JwtController.cs
ibede: According to the OpenID [specification](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) the `/.well-known` paths should be available right at the root of the Issuer. So I think if we want to provide the OpenID info then we have to configure Caddy to handle those paths as well. I'll have a look.

Crypter.Core/Identity/TokenSettings.cs
Jack-Edwards: I'd like for the version in source control to be updated as well.  This was possible to update before the Volumes directory was added to `Crypter.API/.gitignore`.

I think ignoring most of the Volumes directory is a good idea, but the child files currently in source control are important to maintain.

Crypter.API/Controllers/JwtController.cs
Jack-Edwards: Adding another rule to the Caddyfile in Crypter.Web seems like the best option.

I was considering adding an `api.crypter.dev` subdomain and dropping the "api" route prefix, but I'm afraid of breaking any self-hosted setups that might be out there.

Crypter.Core/Identity/TokenSettings.cs
ibede: First I didn't get what you meant - as I've run the API locally the `Volumes` folder is right in `Crypter.API`, but now I get it as I found the `appsettings.json` in the `Volumes/API` in the repository root. So I fixed it. I also mounted the `keys` folder then so the signer key will persist there then.

Crypter.Web/Caddyfile
```diff
@@ -14,6 +14,14 @@
 		reverse_proxy {$CRYPTER_API_BASE}
 	}
 
+	handle /.well-known/* {
+		request_body {
+			max_size {$CADDY_MAX_REQUEST_BODY}
```
Jack-Edwards, line 19: `max_size` shouldn't be necessary here.  It's only declared in the API handler to allow large file transfers through.

Crypter.API/Program.cs
```diff
@@ -57,32 +58,33 @@
...
+        ?? throw new ConfigurationException("TransferStorageSettings missing from configuration"),
         builder.Configuration.GetConnectionString("DefaultConnection")
-        ?? throw new Exception("DefaultConnection missing from configuration"),
+        ?? throw new ConfigurationException("DefaultConnection missing from configuration"),
         hangfireConnectionString)
     .AddBackgroundServer(builder.Configuration.GetSection("HangfireSettings").Get<HangfireSettings>()
-                         ?? throw new Exception("HangfireSettings missing from configuration"));
+                         ?? throw new ConfigurationException("HangfireSettings missing from configuration"));
+
+builder.Services.AddTransient<IConfigureOptions<JwtBearerOptions>, JwtBearerConfiguration>();
```
Jack-Edwards, line 85: Why Transient over Scoped?

Crypter.Test/Core_Tests/Identity_Tests/EdDsa_Tests.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Text;
+
+namespace Crypter.Test.Core_Tests.Identity_Tests
+{
+    [TestFixture]
+    public class EdDsa_Tests
+    {
+        private const string KeyPathSimple = "./simple.private";
+        private const string KeyPathEncrypted = "./encrypted.private";
+        private const string KeyEncriptionPassword = "Pass123";
```
Jack-Edwards, line 42: Minor typo.  `Encription` => `Encryption`

Crypter.Test/Core_Tests/Identity_Tests/EdDsa_Tests.cs
```diff
@@ -0,0 +1,105 @@
...
+        public void Can_Persist_Encrypted_EdDsa_Private_Key()
+        {
+            EdDsaAlgorithm alg = EdDsaAlgorithm.Create(new DefaultCryptoProvider());
+            Assert.That(alg, Is.Not.Null);
+
+            alg.TryExportPrivateKey(KeyPathEncrypted, KeyEncriptionPassword);
+
+            Assert.That(File.Exists(KeyPathEncrypted));
+            Assert.That(new FileInfo(KeyPathEncrypted).Length > 0);
+            File.Delete(KeyPathEncrypted);
```
Jack-Edwards, line 67: I expect these files will not be deleted in case any of the assertions fail.  Wrap `File.Delete` in a `finally` block?

Crypter.Crypto.Providers.Default/Wrappers/Encryption.cs
```diff
@@ -0,0 +1,104 @@
...
+using Geralt;
+
+namespace Crypter.Crypto.Providers.Default.Wrappers
+{
+    [UnsupportedOSPlatform("browser")]
+    public class Encryption : IEncryption
+    {
+        private readonly IPasswordHash _passwordHash;
+
+        public uint KeySize => ChaCha20Poly1305.KeySize;
```
Jack-Edwards, line 41: `Crypter.Crypto.Providers.Browser` uses `XChaCha20Poly1305`.  I'd like for the two providers to use the same algorithm, that way any non-browser clients that may exist in the future would be interoperable with the browser client.

Jack-Edwards:
Will finish reviewing later this weekend, I hope.

Crypter.Core/Identity/Tokens/EdDsaAlgorithm.cs
```diff
@@ -0,0 +1,118 @@
...
+            }
+            catch (Exception ex)
+            {
+                throw new IOException("Unable to read/process the private key file.", ex);
+            }
+        }
+
+        public void TryExportPrivateKey(string filePath, string? passPhrase)
+        {
+            if (Path.GetDirectoryName(filePath) is string directory) {
```
Jack-Edwards, line 107: Rider suggests using `is { } directory` here.

Crypter.Core/Identity/Tokens/EdDsaAlgorithm.cs
```diff
@@ -0,0 +1,118 @@
...
+                Directory.CreateDirectory(directory);
+            }
+
+            string? encodedPrivateKey = passPhrase != null ?
```
Jack-Edwards, line 111: Seems like this should never be null.  `string?` => `string.`

Crypter.Core/Identity/Tokens/EdDsaAlgorithm.cs
```diff
@@ -0,0 +1,118 @@
...
+        public static EdDsaAlgorithm Create(ICryptoProvider? cryptoProvider)
+        {
+            return cryptoProvider == null ? throw new ArgumentNullException(nameof(cryptoProvider)) : new EdDsaAlgorithm(cryptoProvider);
+        }
+
+        public override string SignatureAlgorithm => Name;
+
+        public override KeySizes[] LegalKeySizes => [new KeySizes(32, 32, 0)];
+
+        public override int KeySize => KeyPair?.PrivateKey?.Length ?? KeyPair?.PublicKey?.Length ?? throw new InvalidOperationException("Missing EdDsa key");
```
Jack-Edwards, line 65: `KeyPair` is not declared as nullable.  The condition `?` operators seem unnecessary.

Crypter.Core/Identity/TokenKeyProvider.cs
```diff
@@ -0,0 +1,87 @@
...
+        public JsonWebKey PublicJWK => _publicKey.AsJWK();
+
+        public TokenKeyProvider(ICrypterCryptoProvider cryptoProvider, TokenSettings settings)
+        {
+            EdDsaAlgorithm? edDsa;
+            if (settings.RequirePersistentSigningKey)
+            {
+                if (settings.SigningKeyGenerationStrategy == SigningKeyGenerationStrategy.Always)
+                {
+                    edDsa = EdDsaAlgorithm.Create(cryptoProvider);
```
Jack-Edwards, line 57: What are your thoughts on generating the EdDsa key from a seed?  https://www.geralt.xyz/digital-signatures#generatekeypair-1

It would simplify the configuration if we only needed to define the seed value in appsettings.

Jack-Edwards (changes requested):
See comments.

Crypter.Crypto.Providers.Default/Wrappers/Encryption.cs
Jack-Edwards: Before this gets released, I will write up some tests to verify interoperability.

Crypter.API/Program.cs
ibede: First I thought it's a good idea to recreate this every time independently from requests for memory-wise security, but actually it does not make sense. Even more I think this can be Singleton as it is used only once and the wrapped `TokenService` is Singleton anyway. Dotnet itself also uses typed `IConfigureOptions` as singletons.

Crypter.Core/Identity/TokenKeyProvider.cs
Jack-Edwards: To clarify this a bit, the change would be we store the seed in appsettings.json, similar to how the symmetric key was stored previously.  Instead of storing password-protected files and having logic/settings to decide when those files are generated and how they are stored.

ibede: I'm afraid, I don't get it yet fully. So instead of persisting the keys, we would regenerate both keys every startup with the configured/same seed? But I think that would generate different random keys anyway, though they would be deterministic having the same seed. That would be the same mechanism as with `SigningKeyGenerationStrategy.Always`. What do I miss?

Jack-Edwards: My understanding is the keys would be the same every time they are generated, for a given seed.  I could be wrong.

ibede: You're right, I just checked it with a unit test. I thought the same seed only meant that the keys could be derived from each other. Well, I was half-right, they can be derived: they are the same. 😄 So in that case the seed is like the passphrase for the encrypted key-file. I think it's fine if we generate it on-the-fly every startup.
So I would then switch to this method, I like it. Should I keep the persisted-key method as well? Maybe someone would use that... I don't know.

Jack-Edwards: I lean towards discarding the code that persists keys.  Keep things simple.

Jack-Edwards: But I'm not completely opposed to keeping the persistent key code around.  I'll leave it up to you - my main concern is keeping the `appsettings` intuitive and organized.

Crypter.Crypto.Common/Encryption/IEncryption.cs
```diff
@@ -33,7 +33,9 @@ public interface IEncryption
     uint KeySize { get; }
     uint NonceSize { get; }
     byte[] Encrypt(ReadOnlySpan<byte> key, ReadOnlySpan<byte> nonce, ReadOnlySpan<byte> plaintext);
-    byte[] Encrypt(ReadOnlySpan<byte> key, ReadOnlySpan<byte> nonce, string plaintext);
-    byte[] Decrypt(ReadOnlySpan<byte> key, ReadOnlySpan<byte> nonce, ReadOnlySpan<byte> ciphertext);
+    byte[] Encrypt(ReadOnlySpan<byte> key, ReadOnlySpan<byte> nonce, string plaintext);
+    byte[] Encrypt(string passPhrase, ReadOnlySpan<byte> plainText);
+    byte[] Decrypt(ReadOnlySpan<byte> key, ReadOnlySpan<byte> nonce, ReadOnlySpan<byte> ciphertext);
+    byte[] Decrypt(string passPhrase, ReadOnlySpan<byte> cipherText);
```
Jack-Edwards, line 39: Can the `Encrypt` and `Decrypt` overloads be removed now that we're no longer encrypting/decrypting the key file on the server?

I like the design of the overloads in principle, but the implicit usage of Argon2id seems risky considering it requires a significant amount of dedicated memory to run.  This is normally fine for client scenarios, especially outside the browser.  But it would put more pressure on servers and might force some people to upgrade if they are running the project out of the cheapest VPS they can find (like me).

If there were a use-case in the latest iteration of the PR then I'd say let's figure something out.

Crypter.Core/Identity/TokenKeyProvider.cs
Jack-Edwards: Thank you for making the change.  Looks good.

Crypter.Crypto.Common/Encryption/IEncryption.cs
Jack-Edwards: Pretty certain this is my final comment.  I'll merge after we close this conversation.

Jack-Edwards (approved):
Great work.  Thank you @ibede

PR #789, opened 2025-03-17
Store un-verified email addresses in UserEmailChange table

### LINKED ISSUE #788: [Feature] Wait for new email address to be verified before clearing the current email address
**Describe the feature**
The verified email address for a user account should not be immediately replaced with the newly requested, un-verified email address.

Rather, store the requested email address in the database separately.  Overwrite the UserEntity.EmailAddress only when the new email address is verified by the user.

**Implementation**

- Remove the 'EmailVerified` column from the `UserEntity` table.
- Rename `UserEmailVerificationEntity` to `UserEmailChangeEntity`.
- Add an `EmailAddress` column to the `UserEmailChangeEntity` table.
- Upon user registration, if the user has provided a valid and available email address, then ensure a row is added to the `UserEmailChangeEntity` table with the requested email address.  The email address should not be added to the `UserEntity` data.
- Upon receiving a request to update a user's contact information, verify include the `UserEmailChangeEntity` table when checking whether the requested email address is available.  Do not delete the email address from the `UserEntity` table _unless_ the requested email address is empty or null.  If the requested email address is different than either the `UserEntity.EmailAddress` or `UserEmailChangeEntity.EmailAddress` then delete whatever row may be in the `UserEmailChangeEntity` table and create a new row with the new email address (this last part is performed via Hangfire background job).
- I'm tempted to suggest refactoring how the `UserEmailChangeEntity` record gets inserted.  Ideally, I think the row should be inserted with the requested email address within the request pipeline.  A Hangfire job should be enqueued for the inserted row in order to populate additional information (such as the verification code) and to subsequently send the verification email.

**Important Migration Note**
- All un-verified email address need to be removed from the `UserEntity` table prior to deleting the `UserEntity.EmailVerified` column.  I am not concerned with moving these rows over to the `UserEmailChangeEntity` table; users will have the opportunity to re-request their email changes.
- Similarly, all rows should be removed from the `UserEmailChangeEntity` table.  These requests are stale anyway.

Jack-Edwards:
Resolve #788

Un-verified email addresses are now stored in the `UserEmailChange` table.  These rows are created immediately, during the request pipeline.  Previously, the rows in the now-dropped `UserEmailVerification` table would have been created by the background service.  The rows are created in the request pipeline to prevent problems arising from multiple users requesting the same email address within a short amount of time.

The background service is still responsible for populating the `UserEmailChange` record with verification information and for sending the verification email.

Verification emails now expire after 30 minutes.  This is accomplished by adding a job to the background service which deletes the record from the `UserEmailChange` table.  This job uses both the `UserId` and the `VerificationCode` as lookup information; the `VerificationCode` is necessary in case the user submits a different email change, which will end up with a different verification code.  I do not want the background job enqueued for the first entry to end up expiring the later entry.

Added a SQL script to clear all un-verified email addresses from the `User` table.  This needs to be run manually, as a pre-deployment step.

Updated NuGet packages.


Jack-Edwards:
Trying to find the time to test this.  Having difficulty building Crypter.Web with Docker at the moment.

Jack-Edwards:
Tested pretty well so far:

- Email addresses are first put in the UserEmailChange table.
- Verifying an email address moves that email address to the User table.  The row is removed from UserEmailChange.
- User registration is prohibited if the provided email address exists in either table.

Bug:

- Changing your existing email address to an email address that is already in the UserEmailChange table does not behave properly.  The database remains the same, but the API is responding 200.  The API should respond 409.


To improve:

- The email change UI is horrible, now.  There is no acknowledgement or UX to indicate things worked (or didn't).

Jack-Edwards:
@ibede Let me know if you have interest in reviewing these changes.

Jack-Edwards (commented):
I should include a script that deletes un-verified email addresses from the `User` table.

PR #798, opened 2025-04-15
Update API to return upload settings based upon user category

Jack-Edwards:
The primary motivation for this change is to move the maximum upload size and the maximum amount of disk space the application will consume from `appsettings.json` to the database.  I can update values in the database without having to restart the application.

In addition, I also want to configure different limits for anonymous users versus authenticated or verified users.  For example, I want to be able to reserve a fraction of the 10 GB available in production for authenticated/verified users.

Todo:

- [x] Add database table
- [x] Add migration
- [x] Add seed data
- [x] Update API to query from new database table
- [x] Enforce quota during message and file uploads
- [x] Update front-end to observe quotas
- [x] Test in staging


Crypter.API/appsettings.json
```diff
@@ -16,7 +16,8 @@
       ]
    },
    "DatabaseSettings": {
-      "MigrateOnStartup": true
+      "MigrateOnStartup": true,
+      "SeedOnStartup": true
```
Jack-Edwards, line 20: Need to undo this addition.  Going to seed the data another way.

Crypter.Common.Client/Crypter.Common.Client.csproj
```diff
@@ -14,7 +14,14 @@
...
 
    <ItemGroup>
-     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="9.0.3" />
+     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="9.0.4" />
+     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="9.0.4" />
+   </ItemGroup>
+
+   <ItemGroup>
+     <Reference Include="Microsoft.Extensions.Caching.Abstractions">
+       <HintPath>..\..\..\.nuget\packages\microsoft.extensions.caching.abstractions\9.0.3\lib\net9.0\Microsoft.Extensions.Caching.Abstractions.dll</HintPath>
```
Jack-Edwards, line 23: wtf

PR #834, opened 2026-07-31
Add monad assertions to the test project

Jack-Edwards:
Adds `Assert.IsSome`, `IsNone`, `IsRight`, `IsLeft`, and `IsNeither` for `Maybe` and `Either`, each with an overload taking an expected value and an `Async` variant taking the task directly. The ones that match a value return it, so a test can assert the state and bind the value in one statement, and the failure messages name the actual state and its payload.

`Assert` here is a subclass of NUnit's, declared in the `Crypter.Test` namespace. Unqualified `Assert` in a test now resolves to it and reaches the inherited assertions by the same name. No existing assertions were changed.

The assertions that return a value throw on the wrong state rather than calling `Fail`. Inside an `Assert.Multiple` scope `Fail` records the failure and returns, which would leave them handing back a default value for the test to bind and dereference. `IsNone` and `IsNeither` call `Fail`, so a multiple scope still collects them.

### DIFF
--- a/Crypter.Test/Assert.cs
+++ b/Crypter.Test/Assert.cs
@@ -48,7 +48,7 @@ internal abstract class Assert : NUnit.Framework.Assert
     {
         if (maybe.IsNone)
         {
-            Fail($"Expected Some, but was {Describe(maybe)}.");
+            throw new AssertionException($"Expected Some, but was {Describe(maybe)}.");
         }
 
         return maybe.SomeOrDefault()!;
@@ -83,7 +83,7 @@ internal abstract class Assert : NUnit.Framework.Assert
     {
         if (!either.IsRight)
         {
-            Fail($"Expected Right, but was {Describe(either)}.");
+            throw new AssertionException($"Expected Right, but was {Describe(either)}.");
         }
 
         return either.RightOrDefault(default!)!;
@@ -107,7 +107,7 @@ internal abstract class Assert : NUnit.Framework.Assert
     {
         if (!either.IsLeft)
         {
-            Fail($"Expected Left, but was {Describe(either)}.");
+            throw new AssertionException($"Expected Left, but was {Describe(either)}.");
         }
 
         return either.LeftOrDefault(default!)!;
--- a/Crypter.Test/Assert.cs
+++ b/Crypter.Test/Assert.cs
@@ -1,0 +1,223 @@
+/*
+ * Copyright (C) 2026 Crypter File Transfer
+ *
+ * This file is part of the Crypter file transfer project.
+ *
+ * Crypter is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * The Crypter source code is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * You can be released from the requirements of the aforementioned license
+ * by purchasing a commercial license. Buying such a license is mandatory
+ * as soon as you develop commercial activities involving the Crypter source
+ * code without disclosing the source code of your own applications.
+ *
+ * Contact the current copyright holder to discuss commercial license options.
+ */
+
+using System.Threading.Tasks;
+using EasyMonads;
+using NUnit.Framework;
+
+namespace Crypter.Test;
+
+/// <summary>
+/// NUnit's assertions plus assertions for <see cref="Maybe{T}"/> and <see cref="Either{TLeft,TRight}"/>.
+/// </summary>
+/// <remarks>
+/// This type shadows <see cref="NUnit.Framework.Assert"/> for every test in the Crypter.Test namespace,
+/// so the inherited assertions are reached through the same <c>Assert</c> name they always were.
+/// The monad assertions return the matched value, which lets a test assert the state and bind the
+/// value in one statement.
+/// </remarks>
+internal abstract class Assert : NUnit.Framework.Assert
+{
+    /// <summary>
+    /// Assert that the maybe is in the Some state and return the value it holds.
+    /// </summary>
+    internal static T IsSome<T>(Maybe<T> maybe)
+    {
+        if (maybe.IsNone)
+        {
+            Fail($"Expected Some, but was {Describe(maybe)}.");
+        }
+
+        return maybe.SomeOrDefault()!;
+    }
+
+    /// <summary>
+    /// Assert that the maybe is in the Some state and holds <paramref name="expectedValue"/>, then
+    /// return the value it holds.
+    /// </summary>
+    internal static T IsSome<T>(Maybe<T> maybe, T expectedValue)
+    {
+        T value = IsSome(maybe);
+        That(value, Is.EqualTo(expectedValue));
+        return value;
+    }
+
+    /// <summary>
+    /// Assert that the maybe is in the None state.
+    /// </summary>
+    internal static void IsNone<T>(Maybe<T> maybe)
+    {
+        if (maybe.IsSome)
+        {
+            Fail($"Expected None, but was {Describe(maybe)}.");
+        }
+    }
+
+    /// <summary>
+    /// Assert that the either is in the Right state and return the value it holds.
+    /// </summary>
+    internal static TRight IsRight<TLeft, TRight>(Either<TLeft, TRight> either)
+    {
+        if (!either.IsRight)
+        {
+            Fail($"Expected Right, but was {Describe(either)}.");
+        }
+
+        return either.RightOrDefault(default!)!;
+    }
+
+    /// <summary>
+    /// Assert that the either is in the Right state and holds <paramref name="expectedValue"/>, then
+    /// return the value it holds.
+    /// </summary>
+    internal static TRight IsRight<TLeft, TRight>(Either<TLeft, TRight> either, TRight expectedValue)
+    {
+        TRight value = IsRight(either);
+        That(value, Is.EqualTo(expectedValue));
+        return value;
+    }
+
+    /// <summary>
+    /// Assert that the either is in the Left state and return the value it holds.
+    /// </summary>
+    internal static TLeft IsLeft<TLeft, TRight>(Either<TLeft, TRight> either)
+    {
+        if (!either.IsLeft)
+        {
+            Fail($"Expected Left, but was {Describe(either)}.");
+        }
+
+        return either.LeftOrDefault(default!)!;
+    }
+
+    /// <summary>
+    /// Assert that the either is in the Left state and holds <paramref name="expectedValue"/>, then
+    /// return the value it holds.
+    /// </summary>
+    internal static TLeft IsLeft<TLeft, TRight>(Either<TLeft, TRight> either, TLeft expectedValue)
+    {
+        TLeft value = IsLeft(either);
+        That(value, Is.EqualTo(expectedValue));
+        return value;
+    }
+
+    /// <summary>
+    /// Assert that the either is in the Neither state.
+    /// </summary>
+    internal static void IsNeither<TLeft, TRight>(Either<TLeft, TRight> either)
+    {
+        if (!either.IsNeither)
+        {
+            Fail($"Expected Neither, but was {Describe(either)}.");
+        }
+    }
+
+    /// <summary>
+    /// Await the task, then assert that the maybe is in the Some state and return the value it holds.
+    /// </summary>
+    internal static async Task<T> IsSomeAsync<T>(Task<Maybe<T>> maybeTask)
+    {
+        return IsSome(await maybeTask);
+    }
+
+    /// <summary>
+    /// Await the task, then assert that the maybe is in the Some state and holds
+    /// <paramref name="expectedValue"/>, then return the value it holds.
+    /// </summary>
+    internal static async Task<T> IsSomeAsync<T>(Task<Maybe<T>> maybeTask, T expectedValue)
+    {
+        return IsSome(await maybeTask, expectedValue);
+    }
+
+    /// <summary>
+    /// Await the task, then assert that the maybe is in the None state.
+    /// </summary>
+    internal static async Task IsNoneAsync<T>(Task<Maybe<T>> maybeTask)
+    {
+        IsNone(await maybeTask);
+    }
+
+    /// <summary>
+    /// Await the task, then assert that the either is in the Right state and return the value it holds.
+    /// </summary>
+    internal static async Task<TRight> IsRightAsync<TLeft, TRight>(Task<Either<TLeft, TRight>> eitherTask)
+    {
+        return IsRight(await eitherTask);
+    }
+
+    /// <summary>
+    /// Await the task, then assert that the either is in the Right state and holds
+    /// <paramref name="expectedValue"/>, then return the value it holds.
+    /// </summary>
+    internal static async Task<TRight> IsRightAsync<TLeft, TRight>(Task<Either<TLeft, TRight>> eitherTask, TRight expectedValue)
+    {
+        return IsRight(await eitherTask, expectedValue);
+    }
+
+    /// <summary>
+    /// Await the task, then assert that the either is in the Left state and return the value it holds.
+    /// </summary>
+    internal static async Task<TLeft> IsLeftAsync<TLeft, TRight>(Task<Either<TLeft, TRight>> eitherTask)
+    {
+        return IsLeft(await eitherTask);
+    }
+
+    /// <summary>
+    /// Await the task, then assert that the either is in the Left state and holds
+    /// <paramref name="expectedValue"/>, then return the value it holds.
+    /// </summary>
+    internal static async Task<TLeft> IsLeftAsync<TLeft, TRight>(Task<Either<TLeft, TRight>> eitherTask, TLeft expectedValue)
+    {
+        return IsLeft(await eitherTask, expectedValue);
+    }
+
+    /// <summary>
+    /// Await the task, then assert that the either is in the Neither state.
+    /// </summary>
+    internal static async Task IsNeitherAsync<TLeft, TRight>(Task<Either<TLeft, TRight>> eitherTask)
+    {
+        IsNeither(await eitherTask);
+    }
+
+    private static string Describe<T>(Maybe<T> maybe)
+    {
+        return maybe.IsSome
+            ? $"Some({maybe.SomeOrDefault()})"
+            : "None";
+    }
+
+    private static string Describe<TLeft, TRight>(Either<TLeft, TRight> either)
+    {
+        if (either.IsRight)
+        {
+            return $"Right({either.RightOrDefault(default!)})";
+        }
+
+        return either.IsLeft
+            ? $"Left({either.LeftOrDefault(default!)})"
+            : "Neither";
+    }
+}
--- a/Crypter.Test/Common_Tests/Assert_Tests.cs
+++ b/Crypter.Test/Common_Tests/Assert_Tests.cs
@@ -110,6 +110,13 @@ internal class Assert_Tests
     }
 
     [Test]
+    public void Is_Right_Fails_For_Neither()
+    {
+        AssertionException? exception = Assert.Throws<AssertionException>(() => Assert.IsRight(Neither));
+        Assert.That(exception!.Message, Is.EqualTo("Expected Right, but was Neither."));
+    }
+
+    [Test]
     public void Is_Right_With_An_Expected_Value_Returns_The_Value()
     {
         string value = Assert.IsRight(Right, SomeValue);
@@ -137,6 +144,13 @@ internal class Assert_Tests
     }
 
     [Test]
+    public void Is_Left_Fails_For_Neither()
+    {
+        AssertionException? exception = Assert.Throws<AssertionException>(() => Assert.IsLeft(Neither));
+        Assert.That(exception!.Message, Is.EqualTo("Expected Left, but was Neither."));
+    }
+
+    [Test]
     public void Is_Left_With_An_Expected_Value_Returns_The_Value()
     {
         int value = Assert.IsLeft(Left, LeftValue);
@@ -160,6 +174,48 @@ internal class Assert_Tests
     {
         AssertionException? exception = Assert.Throws<AssertionException>(() => Assert.IsNeither(Right));
         Assert.That(exception!.Message, Is.EqualTo($"Expected Neither, but was Right({SomeValue})."));
+    }
+
+    [Test]
+    public void Is_Some_Inside_A_Multiple_Scope_Does_Not_Return_A_Value()
+    {
+        bool returned = false;
+
+        Assert.Throws<AssertionException>(() => Assert.Multiple(() =>
+        {
+            _ = Assert.IsSome(None);
+            returned = true;
+        }));
+
+        Assert.That(returned, Is.False);
+    }
+
+    [Test]
+    public void Is_Right_Inside_A_Multiple_Scope_Does_Not_Return_A_Value()
+    {
+        bool returned = false;
+
+        Assert.Throws<AssertionException>(() => Assert.Multiple(() =>
+        {
+            _ = Assert.IsRight(Left);
+            returned = true;
+        }));
+
+        Assert.That(returned, Is.False);
+    }
+
+    [Test]
+    public void Is_Left_Inside_A_Multiple_Scope_Does_Not_Return_A_Value()
+    {
+        bool returned = false;
+
+        Assert.Throws<AssertionException>(() => Assert.Multiple(() =>
+        {
+            _ = Assert.IsLeft(Right);
+            returned = true;
+        }));
+
+        Assert.That(returned, Is.False);
     }
 
     [Test]
--- a/Crypter.Test/Common_Tests/Assert_Tests.cs
+++ b/Crypter.Test/Common_Tests/Assert_Tests.cs
@@ -1,0 +1,276 @@
+/*
+ * Copyright (C) 2026 Crypter File Transfer
+ *
+ * This file is part of the Crypter file transfer project.
+ *
+ * Crypter is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * The Crypter source code is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * You can be released from the requirements of the aforementioned license
+ * by purchasing a commercial license. Buying such a license is mandatory
+ * as soon as you develop commercial activities involving the Crypter source
+ * code without disclosing the source code of your own applications.
+ *
+ * Contact the current copyright holder to discuss commercial license options.
+ */
+
+using System.Threading.Tasks;
+using EasyMonads;
+using NUnit.Framework;
+
+namespace Crypter.Test.Common_Tests;
+
+/// <summary>
+/// Every monad assertion is exercised twice: once against the state it expects, and once against a
+/// state it must reject. A helper that never fails would silently pass every test that depends on it.
+/// </summary>
+/// <remarks>
+/// The rejection tests wrap the assertion in <c>Assert.Throws</c> rather than a try/catch. NUnit
+/// records a failed assertion against the running test as soon as it happens, and only
+/// <c>Assert.Throws</c> clears that record. A caught <see cref="AssertionException"/> still fails the
+/// test.
+/// </remarks>
+[TestFixture]
+internal class Assert_Tests
+{
+    private const string SomeValue = "Frodo";
+    private const string OtherValue = "Bilbo";
+    private const int LeftValue = 42;
+    private const int OtherLeftValue = 24;
+
+    private static Maybe<string> Some => Maybe<string>.From(SomeValue);
+    private static Maybe<string> None => Maybe<string>.None;
+
+    private static Either<int, string> Right => Either<int, string>.FromRight(SomeValue);
+    private static Either<int, string> Left => Either<int, string>.FromLeft(LeftValue);
+    private static Either<int, string> Neither => Either<int, string>.Neither;
+
+    [Test]
+    public void Is_Some_Returns_The_Value()
+    {
+        string value = Assert.IsSome(Some);
+        Assert.That(value, Is.EqualTo(SomeValue));
+    }
+
+    [Test]
+    public void Is_Some_Fails_For_None()
+    {
+        AssertionException? exception = Assert.Throws<AssertionException>(() => Assert.IsSome(None));
+        Assert.That(exception!.Message, Is.EqualTo("Expected Some, but was None."));
+    }
+
+    [Test]
+    public void Is_Some_With_An_Expected_Value_Returns_The_Value()
+    {
+        string value = Assert.IsSome(Some, SomeValue);
+        Assert.That(value, Is.EqualTo(SomeValue));
+    }
+
+    [Test]
+    public void Is_Some_With_An_Expected_Value_Fails_For_A_Different_Value()
+    {
+        Assert.Throws<AssertionException>(() => Assert.IsSome(Some, OtherValue));
+    }
+
+    [Test]
+    public void Is_None_Passes_For_None()
+    {
+        Assert.IsNone(None);
+    }
+
+    [Test]
+    public void Is_None_Fails_For_Some()
+    {
+        AssertionException? exception = Assert.Throws<AssertionException>(() => Assert.IsNone(Some));
+        Assert.That(exception!.Message, Is.EqualTo($"Expected None, but was Some({SomeValue})."));
+    }
+
+    [Test]
+    public void Is_Right_Returns_The_Value()
+    {
+        string value = Assert.IsRight(Right);
+        Assert.That(value, Is.EqualTo(SomeValue));
+    }
+
+    [Test]
+    public void Is_Right_Fails_For_Left()
+    {
+        AssertionException? exception = Assert.Throws<AssertionException>(() => Assert.IsRight(Left));
+        Assert.That(exception!.Message, Is.EqualTo($"Expected Right, but was Left({LeftValue})."));
+    }
+
+    [Test]
+    public void Is_Right_With_An_Expected_Value_Returns_The_Value()
+    {
+        string value = Assert.IsRight(Right, SomeValue);
+        Assert.That(value, Is.EqualTo(SomeValue));
+    }
+
+    [Test]
+    public void Is_Right_With_An_Expected_Value_Fails_For_A_Different_Value()
+    {
+        Assert.Throws<AssertionException>(() => Assert.IsRight(Right, OtherValue));
+    }
+
+    [Test]
+    public void Is_Left_Returns_The_Value()
+    {
+        int value = Assert.IsLeft(Left);
+        Assert.That(value, Is.EqualTo(LeftValue));
+    }
+
+    [Test]
+    public void Is_Left_Fails_For_Right()
+    {
+        AssertionException? exception = Assert.Throws<AssertionException>(() => Assert.IsLeft(Right));
+        Assert.That(exception!.Message, Is.EqualTo($"Expected Left, but was Right({SomeValue})."));
+    }
+
+    [Test]
+    public void Is_Left_With_An_Expected_Value_Returns_The_Value()
+    {
+        int value = Assert.IsLeft(Left, LeftValue);
+        Assert.That(value, Is.EqualTo(LeftValue));
+    }
+
+    [Test]
+    public void Is_Left_With_An_Expected_Value_Fails_For_A_Different_Value()
+    {
+        Assert.Throws<AssertionException>(() => Assert.IsLeft(Left, OtherLeftValue));
+    }
+
+    [Test]
+    public void Is_Neither_Passes_For_Neither()
+    {
+        Assert.IsNeither(Neither);
+    }
+
+    [Test]
+    public void Is_Neither_Fails_For_Right()
+    {
+        AssertionException? exception = Assert.Throws<AssertionException>(() => Assert.IsNeither(Right));
+        Assert.That(exception!.Message, Is.EqualTo($"Expected Neither, but was Right({SomeValue})."));
+    }
+
+    [Test]
+    public async Task Is_Some_Async_Returns_The_Value_Async()
+    {
+        string value = await Assert.IsSomeAsync(Task.FromResult(Some));
+        Assert.That(value, Is.EqualTo(SomeValue));
+    }
+
+    [Test]
+    public void Is_Some_Async_Fails_For_None()
+    {
+        AssertionException? exception =
+            Assert.ThrowsAsync<AssertionException>(() => Assert.IsSomeAsync(Task.FromResult(None)));
+        Assert.That(exception!.Message, Is.EqualTo("Expected Some, but was None."));
+    }
+
+    [Test]
+    public async Task Is_Some_Async_With_An_Expected_Value_Returns_The_Value_Async()
+    {
+        string value = await Assert.IsSomeAsync(Task.FromResult(Some), SomeValue);
+        Assert.That(value, Is.EqualTo(SomeValue));
+    }
+
+    [Test]
+    public void Is_Some_Async_With_An_Expected_Value_Fails_For_A_Different_Value()
+    {
+        Assert.ThrowsAsync<AssertionException>(() => Assert.IsSomeAsync(Task.FromResult(Some), OtherValue));
+    }
+
+    [Test]
+    public async Task Is_None_Async_Passes_For_None_Async()
+    {
+        await Assert.IsNoneAsync(Task.FromResult(None));
+    }
+
+    [Test]
+    public void Is_None_Async_Fails_For_Some()
+    {
+        AssertionException? exception =
+            Assert.ThrowsAsync<AssertionException>(() => Assert.IsNoneAsync(Task.FromResult(Some)));
+        Assert.That(exception!.Message, Is.EqualTo($"Expected None, but was Some({SomeValue})."));
+    }
+
+    [Test]
+    public async Task Is_Right_Async_Returns_The_Value_Async()
+    {
+        string value = await Assert.IsRightAsync(Task.FromResult(Right));
+        Assert.That(value, Is.EqualTo(SomeValue));
+    }
+
+    [Test]
+    public void Is_Right_Async_Fails_For_Left()
+    {
+        AssertionException? exception =
+            Assert.ThrowsAsync<AssertionException>(() => Assert.IsRightAsync(Task.FromResult(Left)));
+        Assert.That(exception!.Message, Is.EqualTo($"Expected Right, but was Left({LeftValue})."));
+    }
+
+    [Test]
+    public async Task Is_Right_Async_With_An_Expected_Value_Returns_The_Value_Async()
+    {
+        string value = await Assert.IsRightAsync(Task.FromResult(Right), SomeValue);
+        Assert.That(value, Is.EqualTo(SomeValue));
+    }
+
+    [Test]
+    public void Is_Right_Async_With_An_Expected_Value_Fails_For_A_Different_Value()
+    {
+        Assert.ThrowsAsync<AssertionException>(() => Assert.IsRightAsync(Task.FromResult(Right), OtherValue));
+    }
+
+    [Test]
+    public async Task Is_Left_Async_Returns_The_Value_Async()
+    {
+        int value = await Assert.IsLeftAsync(Task.FromResult(Left));
+        Assert.That(value, Is.EqualTo(LeftValue));
+    }
+
+    [Test]
+    public void Is_Left_Async_Fails_For_Right()
+    {
+        AssertionException? exception =
+            Assert.ThrowsAsync<AssertionException>(() => Assert.IsLeftAsync(Task.FromResult(Right)));
+        Assert.That(exception!.Message, Is.EqualTo($"Expected Left, but was Right({SomeValue})."));
+    }
+
+    [Test]
+    public async Task Is_Left_Async_With_An_Expected_Value_Returns_The_Value_Async()
+    {
+        int value = await Assert.IsLeftAsync(Task.FromResult(Left), LeftValue);
+        Assert.That(value, Is.EqualTo(LeftValue));
+    }
+
+    [Test]
+    public void Is_Left_Async_With_An_Expected_Value_Fails_For_A_Different_Value()
+    {
+        Assert.ThrowsAsync<AssertionException>(() => Assert.IsLeftAsync(Task.FromResult(Left), OtherLeftValue));
+    }
+
+    [Test]
+    public async Task Is_Neither_Async_Passes_For_Neither_Async()
+    {
+        await Assert.IsNeitherAsync(Task.FromResult(Neither));
+    }
+
+    [Test]
+    public void Is_Neither_Async_Fails_For_Left()
+    {
+        AssertionException? exception =
+            Assert.ThrowsAsync<AssertionException>(() => Assert.IsNeitherAsync(Task.FromResult(Left)));
+        Assert.That(exception!.Message, Is.EqualTo($"Expected Neither, but was Left({LeftValue})."));
+    }
+}


Crypter.Test/Assert.cs
```diff
@@ -0,0 +1,223 @@
...
+
+namespace Crypter.Test;
+
+/// <summary>
+/// NUnit's assertions plus assertions for <see cref="Maybe{T}"/> and <see cref="Either{TLeft,TRight}"/>.
+/// </summary>
+/// <remarks>
+/// This type shadows <see cref="NUnit.Framework.Assert"/> for every test in the Crypter.Test namespace,
+/// so the inherited assertions are reached through the same <c>Assert</c> name they always were.
+/// The monad assertions return the matched value, which lets a test assert the state and bind the
```
Jack-Edwards: **comments-01** · comments lens · verifier: holds

#### Where it holds

`Crypter.Test/Assert.cs:36`–`:41`. The class `<remarks>` documents the shadowing and the
value-returning design but omits the contract a caller actually needs: inside `Assert.Multiple`,
`IsSome`, `IsRight`, `IsLeft` and their async forms throw on failure and end the scope, so nothing
after them in the block is evaluated, while `IsNone`, `IsNeither` and their async forms report
through `Fail` and let the scope continue. A test author mixing the two in one `Assert.Multiple`
block silently loses every check after the first value-returning failure.

The generalisation "The monad assertions return the matched value" is also wrong for the five
void members and should be scoped to the ones that do return.

Fix as the claim directs: state the split in the `<remarks>`, alongside or in place of the
returning-the-value sentence.

<details><summary>Evidence</summary>

`Crypter.Test/Assert.cs` is added whole by this diff (`git diff origin/stable...HEAD --stat`:
`Assert.cs` +223, `Common_Tests/Assert_Tests.cs` +332, nothing else), so the comments in it are
the diff's own.

The two failure mechanisms are where the claim says they are:

- `Assert.cs:51` (`IsSome`), `:86` (`IsRight`), `:110` (`IsLeft`) — `throw new AssertionException(...)`.
- `Assert.cs:75` (`IsNone`), `:134` (`IsNeither`) — `Fail(...)`, i.e. `NUnit.Framework.Assert.Fail`
  via the base class.

The async members inherit the split unchanged: `IsSomeAsync`/`IsRightAsync`/`IsLeftAsync`
(`:141`–`:195`) delegate to the throwing overloads, `IsNoneAsync` (`:158`) and `IsNeitherAsync`
(`:200`) to the `Fail` ones.

The class `<remarks>` is `Assert.cs:36`–`:41`. It covers the shadowing and says "The monad
assertions return the matched value, which lets a test assert the state and bind the value in one
statement." It says nothing about multiple-assert scopes. None of the sixteen member `<summary>`
blocks mentions failure behaviour either — every one of them describes only what is asserted and
what is returned.

The divergence is deliberate and already pinned: `Assert_Tests.cs:179`–`:218` holds
`Is_Some_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` and its `Is_Right`/`Is_Left` siblings,
each asserting that the statement *after* the assertion inside `Assert.Multiple` never executes.
There is no corresponding test for `IsNone`/`IsNeither`, because for those the block does continue.

I compiled this file's `Assert` against NUnit 4.6.1 and EasyMonads 2.5.0 (the versions in
`Crypter.Test.csproj` and `Crypter.Common.csproj`) in a scratch project and ran the claim's
scenario verbatim — `Assert.Multiple(() => { Assert.IsNone(a); string b = Assert.IsSome(x);
thirdReached = true; Assert.IsNone(c); })` with `a` None, `x` None and `c` Some("leaked"):

```
THIRD_REACHED=False
THROWN=AssertionException
MSG=Expected Some, but was None.
```

The `IsSome` failure escapes `Assert.Multiple` as a bare `AssertionException`, the `IsNone(c)`
check never runs, and the Some("leaked") defect is never reported.

The contrast run — two failing void assertions in one block, `Assert.IsNone(some)` then
`Assert.IsNeither(right)`:

```
SECOND_REACHED=True
THROWN2=MultipleAssertException
MSG2=Multiple failures or warnings in test:
  1) Expected None, but was Some(one).   at Crypter.Test.Assert.IsNone[T](...) Assert.cs:line 75
  2) Expected Neither, but was Right(two).  at Crypter.Test.Assert.IsNeither[...](...) Assert.cs:line 134
```

Both failures are recorded and the block runs to the end. A direct NUnit probe confirms the
mechanism is NUnit's and not something this file does: `Assert.Fail` inside `Assert.Multiple`
records and returns (two successive `Fail` calls both appear in the resulting
`MultipleAssertException`, and code between them executes), while a raw `throw new
AssertionException` inside the same scope unwinds it immediately.

So every link the claim describes is there: the two mechanisms sit side by side in one file, they
behave differently in the scope NUnit users reach for when they want several checks reported
together, the difference is invisible at the call site, and the only comment that touches failure
behaviour is both silent on it and — read literally, "the monad assertions return the matched
value" — false for the five members that return nothing (`IsNone`, `IsNeither`, `IsNoneAsync`,
`IsNeitherAsync`).

</details>

Crypter.Test/Common_Tests/Assert_Tests.cs
```diff
@@ -0,0 +1,332 @@
...
+ */
+
+using System.Threading.Tasks;
+using EasyMonads;
+using NUnit.Framework;
+
+namespace Crypter.Test.Common_Tests;
+
+/// <summary>
+/// Every monad assertion is exercised twice: once against the state it expects, and once against a
```
Jack-Edwards: **comments-03** · comments lens · verifier: holds

#### What is broken

`Crypter.Test/Common_Tests/Assert_Tests.cs:34` states a count the fixture does not keep. The
`<summary>` presents itself as the fixture's convention — "Every monad assertion is exercised twice"
with the two cases enumerated after the colon — while five of the 37 tests fall outside that
enumeration, including the entire three-test `Assert.Multiple` category. A maintainer adding an
assertion to `Assert.cs` and following the documented convention writes a pass test and one
rejection test, and never learns that a value-returning assertion also needs the multiple-scope
test that motivates its `throw`.

The claim's proposed direction is the right one: drop the number, keep "A helper that never fails
would silently pass every test that depends on it" (line 35), which is the rationale and is not
readable off the code, and name the multiple-scope case if the summary is going to describe the
fixture's structure.

#### Two qualifications for whoever receives this

The claim is right about the defect and loose in two places, neither of which sinks it:

1. It says the multiple-scope tests "cover a third thing that is neither the expected state nor a
   rejected one." The states those tests use *are* rejected states — `None` (line 186), `Left`
   (200), `Right` (214). What is third is the *property* under test: that no value is returned when
   the assertion fails inside `Assert.Multiple`. The tests are outside the summary's enumeration
   because of the property, not the state.
2. The `IsNeither` half of the failure scenario overstates. `Is_Neither_Fails_For_Right` (173) and
   `Is_Neither_Async_Fails_For_Left` (326) confirm the asymmetry the claim describes — sync
   `IsNeither` is never given a `Left`, async `IsNeither` is never given a `Right` — and the same
   gap exists for the async `IsRight`/`IsLeft` overloads, which are never given `Neither`. But the
   summary promises only *one* rejection state per assertion, so a reader taking it literally would
   not conclude the missing cases are covered. That asymmetry is a coverage observation in its own
   right; it is not something this sentence misleads anyone about.

The verdict rests on point (1) above and the arithmetic: the summary describes a shape 4 of 16
overloads and one whole test category do not have.

<details><summary>The finding and evidence</summary>

> The fixture summary claims a shape the fixture does not have. "Every monad assertion is exercised
> twice: once against the state it expects, and once against a state it must reject" is contradicted
> by the tests directly beneath it: `IsRight` has three (`Is_Right_Returns_The_Value`,
> `Is_Right_Fails_For_Left`, `Is_Right_Fails_For_Neither`), `IsLeft` has three, and three further
> tests — `Is_Some_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` and its two siblings — cover a
> third thing that is neither the expected state nor a rejected one. Counting the multiple-scope
> tests, `IsSome`, `IsRight` and `IsLeft` are each exercised three or four times.
>
> The count is also a number nothing enforces, so it is stale on the next edit. The sentence that
> earns its place is the second one — "A helper that never fails would silently pass every test that
> depends on it" — which is the reason the rejection tests exist and cannot be read off the code.
>
> Failure: a maintainer adding an assertion to `Assert.cs` reads the summary as the fixture's
> convention and writes exactly two tests, matching a rule the file does not follow; one auditing
> coverage of `IsNeither`, which has a sync rejection test for Right and an async one for Left but
> neither the other way round, trusts "every ... twice" and stops looking.
>
> Direction: keep the reason and drop the count — say that each assertion is tested in both
> directions, passing and failing, and let the tests carry how many that takes. The multiple-scope
> tests are a third case worth naming if the summary is going to describe the fixture's structure at
> all.

Both files are new in `origin/stable...HEAD`, so the summary is diff-touched code.

`Crypter.Test/Assert.cs` declares 16 `internal static` assertion overloads (`grep -c "internal
static"` = 16; the two `Describe` helpers are `private`). `Crypter.Test/Common_Tests/Assert_Tests.cs`
carries 37 `[Test]` methods (`grep -c "\[Test\]"` = 37). If the summary at line 34 were true of the
fixture, the count would be 32. The five surplus tests are exactly the ones the claim names:

- `Is_Right_Fails_For_Neither` (Assert_Tests.cs:113) — a second rejection state for `IsRight`
- `Is_Left_Fails_For_Neither` (Assert_Tests.cs:147) — a second rejection state for `IsLeft`
- `Is_Some_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` (Assert_Tests.cs:180)
- `Is_Right_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` (Assert_Tests.cs:194)
- `Is_Left_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` (Assert_Tests.cs:208)

Per-overload tally, reading the file top to bottom: `IsSome(Maybe)` 3 (lines 59, 66, 180),
`IsRight(Either)` 4 (99, 106, 113, 194), `IsLeft(Either)` 4 (133, 140, 147, 208). The remaining 13
overloads — both expected-value overloads of each, `IsNone`, `IsNeither`, and all eight async
overloads — have exactly 2 each. So 12 of 16 overloads match the stated shape and 4 do not, and the
claim's "three or four times" for `IsSome`, `IsRight` and `IsLeft` is accurate as written.

The multiple-scope tests are the substantive omission. They assert that `Assert.IsSome`/`IsRight`/
`IsLeft` do not return a value when the assertion fails inside an `Assert.Multiple` scope — the
reason `Assert.cs` throws `AssertionException` directly for the value-returning assertions
(Assert.cs:50, 85, 108) while the void ones call `Fail` (Assert.cs:73, 130). That is the least
obvious thing in the fixture, and the summary's two-part enumeration has no room for it.

</details>

Crypter.Test/Assert.cs
```diff
@@ -0,0 +1,223 @@
...
+internal abstract class Assert : NUnit.Framework.Assert
+{
+    /// <summary>
+    /// Assert that the maybe is in the Some state and return the value it holds.
+    /// </summary>
+    internal static T IsSome<T>(Maybe<T> maybe)
+    {
+        if (maybe.IsNone)
+        {
+            throw new AssertionException($"Expected Some, but was {Describe(maybe)}.");
```
Jack-Edwards: **correctness-01** · correctness lens · verifier: holds

#### Where it holds

Every link traces, so it holds — with its scope stated precisely, because the finding is narrower
than its opening sentence suggests and whoever fixes it should not over-correct.

What is *not* broken: the throw itself. A method that must return a `T` cannot record-and-continue,
and the three tests above pin that. No test passes when it should fail, and no result is wrong.

What is broken: `Crypter.Test/Assert.cs:51` (and 86, 110) introduces a shared test helper whose
behaviour inside `Assert.Multiple` is the opposite of its siblings' in the same class, and nothing
on the API surface says so. The `<remarks>` at lines 36–41 advertise exactly the affordance that
triggers the loss — "The monad assertions return the matched value, which lets a test assert the
state and bind the value in one statement" — while saying nothing about what that costs in a
`Multiple` block. The concrete failure: a developer writes a `Multiple` block with `That`
assertions ahead of an `Assert.IsSome`, the maybe is `None`, and the run discards the messages of
the earlier failures; they fix the `None`, rerun, and only then learn about the second defect —
the round-trip `Assert.Multiple` exists to prevent.

Both remedies the finding suggests are sound and either one closes it:

- State it in the class `<remarks>`, so a caller knows to place `IsSome`/`IsRight`/`IsLeft` before
  other assertions in a `Multiple` block; or
- add a void-returning overload that uses `Fail(...)` and composes correctly, reserving the
  throwing form for when the value is actually bound.

The first is the smaller change and matches the finding's framing that the design is right and the
documentation is what is missing.

<details><summary>The finding and evidence</summary>

> `IsSome`, `IsRight` and `IsLeft` signal failure by throwing `AssertionException` directly instead
> of going through NUnit's `Fail`. Inside an `Assert.Multiple` block that exception escapes the
> block, and NUnit then reports *only* that exception's message — the messages of every assertion
> that already failed earlier in the same block are discarded.
>
> [...] Note the asymmetry this creates within the same class: `IsNone` and `IsNeither` call
> `Fail(...)`, which records and lets the block continue, so they compose with `Assert.Multiple`
> correctly. Only the three value-returning assertions destroy the report.
>
> The three `_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` tests pin the aborting behaviour as
> intended [...] What is missing is that the cost is neither documented nor bounded.

The code is as described. `Crypter.Test/Assert.cs` is added whole by this diff (223 of the diff's
555 added lines), so every line below is new:

- `IsSome` line 51 — `throw new AssertionException($"Expected Some, but was {Describe(maybe)}.");`
- `IsRight` line 86 — `throw new AssertionException($"Expected Right, but was {Describe(either)}.");`
- `IsLeft` line 110 — `throw new AssertionException($"Expected Left, but was {Describe(either)}.");`

against the sibling assertions that record instead of throwing:

- `IsNone` line 75 — `Fail($"Expected None, but was {Describe(maybe)}.");`
- `IsNeither` line 134 — `Fail($"Expected Neither, but was {Describe(either)}.");`

The asymmetry the finding names is therefore real and is internal to one new class. The
`Async` wrappers (lines 141–203) delegate to these same six methods, so they inherit it.

I did not settle the NUnit behaviour from memory. I built a standalone project pinned to the same
versions this repo pins (`NUnit` 4.6.1, `NUnit3TestAdapter` 6.2.0, `Microsoft.NET.Test.Sdk`
18.8.1, `net10.0` — see `Crypter.Test/Crypter.Test.csproj`), with a helper shaped like this one:
one method that throws `AssertionException` and one that calls `Fail`, each preceded inside an
`Assert.Multiple` block by a `That` that fails first.

The `Fail(...)` path reports both failures in full:

```
Failed FailAfterPendingFailure
  Error Message:
   Multiple failures or warnings in test:
  1)   size mismatch
Assert.That(512, Is.EqualTo(1024))
  Expected: 1024
  But was:  512
  2) Expected None, but was Some(x).
```

The throwing path reports only the thrown message:

```
Failed ThrowAfterPendingFailure
  Error Message:
   Expected Some, but was None.
  Stack Trace:
     at Multi.MyAssert.IsSome[T](T value) ...
1)    at Multi.ThrowInsideMultiple.<>c.<ThrowAfterPendingFailure>b__0_0() in Tests.cs:line 33
```

The earlier failure survives only as the bare numbered stack frame `1)`. Its
`Expected: 1024 / But was: 512` text is gone. This is the finding's predicted report shape, down
to the detail that the lost assertion leaves a frame but no message.

The mechanism is `Assert.Multiple` decrementing `MultipleAssertLevel` in a `finally` and then
throwing `MultipleAssertException` only on normal return; an exception escaping the delegate skips
that aggregation, and the escaping `AssertionException` becomes the result's message on its own.
The finding's own note is correct on this point: recording through `Fail(...)` *and then* throwing
would not help, because the loss comes from the exception escaping `Multiple`, not from bypassing
`Fail`.

The finding's reading of the pinning tests is also accurate.
`Crypter.Test/Common_Tests/Assert_Tests.cs:180-219` asserts `Assert.Throws<AssertionException>`
around a `Multiple` block and that the statement after the assertion never ran, for each of
`IsSome`, `IsRight` and `IsLeft`. The abort is deliberate and specified.

</details>

Crypter.Test/Assert.cs
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        return maybe.SomeOrDefault()!;
+    }
+
+    /// <summary>
+    /// Assert that the maybe is in the Some state and holds <paramref name="expectedValue"/>, then
+    /// return the value it holds.
+    /// </summary>
+    internal static T IsSome<T>(Maybe<T> maybe, T expectedValue)
```
Jack-Edwards: **coverage-01** · coverage lens · verifier: holds

#### What is broken

`Crypter.Test/Assert.cs:61` (and `:96`, `:120`) return a value on failure inside an
`Assert.Multiple` scope, and no test pins that either way.

This is the same failure mode the HEAD commit set out to remove. Its message says "Inside an
Assert.Multiple scope, Fail records the failure and returns rather than throwing. The assertions
that return a value went on to hand back a default […] There is no value to return once the
state is wrong, so these now throw." The state checks now do throw. The value checks do not: they
still record-and-return, so a test that writes

```csharp
Assert.Multiple(() =>
{
    string name = Assert.IsSome(maybe, "Bilbo");
    Assert.That(name.Length, Is.EqualTo(5));
});
```

gets `name` bound to whatever the maybe actually held and every following assertion in the scope
evaluated against it. The diff's own three tests exist to rule that out for the sibling
overloads; the expected-value overloads do the opposite and nothing records whether that is
deliberate.

To fix: decide the intent and pin it. If a value-returning assertion must never return on
failure, replace the `That` call in each expected-value overload with an explicit comparison and
`throw new AssertionException(...)`, and add a `Multiple`-scope test per overload alongside the
existing three. If collecting the mismatch is intended, the tests still need to exist and must
expect `Assert.Throws<MultipleAssertException>` rather than `AssertionException`.

<details><summary>The finding and evidence</summary>

> The three `Assert.Multiple` tests the diff added cover only the single-argument
> overloads. The expected-value overloads — `IsSome(maybe, expectedValue)`,
> `IsRight(either, expectedValue)`, `IsLeft(either, expectedValue)` — fail through
> `That(value, Is.EqualTo(expectedValue))` instead of a `throw`, and no test in
> `Assert_Tests` exercises any of them inside a `Multiple` scope.
>
> That path behaves the opposite way to the one the added tests pin. […] inside
> a `Multiple` scope the assertion records the failure, hands the caller the value
> that failed its own expectation, and lets the rest of the block run on it — which
> is exactly the outcome `Is_Some_Inside_A_Multiple_Scope_Does_Not_Return_A_Value`,
> `Is_Right_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` and
> `Is_Left_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` were written to rule
> out for the sibling overloads.

Both files are added whole by this diff (`git diff origin/stable...HEAD --stat`:
`Crypter.Test/Assert.cs` +223, `Crypter.Test/Common_Tests/Assert_Tests.cs` +332), so
nothing here is pre-existing code the finding has wandered into.

**The two shapes are as described.** In `Crypter.Test/Assert.cs`, the single-argument
overloads throw on a bad state — `IsSome` line 51, `IsRight` line 86, `IsLeft` line 110 all
`throw new AssertionException(...)`. The expected-value overloads delegate the value check to
`That`:

- `IsSome<T>(Maybe<T> maybe, T expectedValue)` — lines 61-66: `IsSome(maybe)`, then
  `That(value, Is.EqualTo(expectedValue))`, then `return value`.
- `IsRight<TLeft, TRight>(Either<TLeft, TRight> either, TRight expectedValue)` — lines 96-101,
  same shape.
- `IsLeft<TLeft, TRight>(Either<TLeft, TRight> either, TLeft expectedValue)` — lines 120-125,
  same shape.

**The coverage gap is as described.** The three `Multiple`-scope tests in
`Crypter.Test/Common_Tests/Assert_Tests.cs` are at lines 179-219, and each one calls a
single-argument overload: `Assert.IsSome(None)` (line 185), `Assert.IsRight(Left)` (line 199),
`Assert.IsLeft(Right)` (line 213). The expected-value overloads are exercised only outside a
`Multiple` scope, by `Is_Some_With_An_Expected_Value_Fails_For_A_Different_Value` (line 80) and
its `IsRight`/`IsLeft` siblings (lines 127, 161), where `That` does throw `AssertionException`
and those tests pass. No test in the file puts an expected-value overload inside `Multiple`.

**The behaviour claim reproduces.** I built a standalone NUnit 4.6.1 / net10.0 probe (the
Crypter.Test fixture needs Docker for Testcontainers, which is unavailable in this container, so
the probe mirrors the `IsSome(maybe, expectedValue)` shape rather than calling it) and ran it:

```
PROBE returned=True captured=Frodo escaped=NUnit.Framework.MultipleAssertException
PROBE MultipleAssertException base chain: MultipleAssertException -> ResultStateException -> Exception -> Object
PROBE AssertionException base chain: AssertionException -> ResultStateException -> Exception -> Object
PROBE isAssertionException=False
```

So inside `Multiple`, a failing `That` records the failure and returns; the assertion hands back
the value that failed its own expectation (`Frodo`, where `Bilbo` was expected) and the rest of
the block runs on it. The exception that escapes is `MultipleAssertException`, which shares only
`ResultStateException` with `AssertionException` — the finding's note about the test expectation
is right.

</details>

Crypter.Test/Assert.cs
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    /// <summary>
+    /// Assert that the maybe is in the None state.
+    /// </summary>
+    internal static void IsNone<T>(Maybe<T> maybe)
+    {
+        if (maybe.IsSome)
+        {
+            Fail($"Expected None, but was {Describe(maybe)}.");
```
Jack-Edwards: **coverage-02** · coverage lens · verifier: holds

#### Where it holds

The diff introduces a deliberate behavioural split and tests only one side of it.
`IsNone` (`Assert.cs:75`) and `IsNeither` (`Assert.cs:134`) use `Fail` so that a
failure inside `Assert.Multiple` is recorded while the block continues. No test
in `Assert_Tests` enters a `Multiple` scope with either one, so nothing pins that
behaviour down: either call site can be changed to `throw new AssertionException`
— collapsing the distinction the commit was written to create — and the suite
stays fully green at 37 passed.

The gap is specific rather than general. The fixture is otherwise thorough, and
the three `Multiple`-scope tests show the author knew this axis mattered; they
were simply written only for the assertions that return a value.

To close it, add one `Multiple`-scope test per assertion, mirroring the existing
three but inverted — asserting that the statement after the failed assertion
*does* run:

```csharp
[Test]
public void Is_None_Inside_A_Multiple_Scope_Records_The_Failure_And_Continues()
{
    bool continued = false;

    Assert.Throws<MultipleAssertException>(() => Assert.Multiple(() =>
    {
        Assert.IsNone(Some);
        continued = true;
    }));

    Assert.That(continued, Is.True);
}
```

and the same shape for `IsNeither` with `Assert.IsNeither(Right)`. As the claim
notes, these must expect `MultipleAssertException`, not `AssertionException` —
confirmed by the probe above. Both were executed in this worktree in exactly the
form shown: against HEAD they pass (`Failed: 0, Passed: 2`), and against the
`throw` mutant they fail (`Failed: 2, Passed: 0`). They close the gap they are
meant to close.

<details><summary>The finding and evidence</summary>

> `IsNone` and `IsNeither` are the two assertions that report failure with `Fail`
> rather than `throw` — the asymmetry commit a8d331ef ("Throw from the monad
> assertions that return a value") introduced deliberately. The behaviour that
> asymmetry exists for is that inside an `Assert.Multiple` scope these two record a
> failure and let the block continue, and no test reaches that state. Both
> `Is_None_Fails_For_Some` and `Is_Neither_Fails_For_Right` call the assertion
> outside a `Multiple` scope, where `Fail` throws `AssertionException` immediately
> and is indistinguishable from an explicit throw.
>
> Verified in this worktree by rewriting both `Fail` calls as
> `throw new AssertionException(...)` and running the fixture: `Passed! - Failed:
> 0, Passed: 37`. The whole `Assert_Tests` suite stays green under a change that
> erases the distinction the diff was built around. A probe confirms the state that
> is going untested — `Assert.IsNone(Maybe<string>.From("Frodo"))` inside
> `Assert.Multiple` returns normally, the block runs on, and `Multiple` ends by
> throwing `MultipleAssertException`.
>
> Direction: add a `Multiple`-scope test for each of `IsNone` and `IsNeither`,
> mirroring the three the diff already added for the value-returning assertions but
> asserting the opposite — that the statement after the failed assertion still
> runs. Those tests have to expect `MultipleAssertException` rather than
> `AssertionException`, since the failure is recorded rather than thrown.

The code is as described. `Crypter.Test/Assert.cs` is added whole by this diff
(`git diff origin/stable...HEAD --stat` shows only `Assert.cs` +223 and
`Assert_Tests.cs` +332), so this is the diff's own code, not surrounding code.

Exactly two assertions report failure through `Fail`:

- `Assert.cs:75` — `IsNone`: `Fail($"Expected None, but was {Describe(maybe)}.");`
- `Assert.cs:134` — `IsNeither`: `Fail($"Expected Neither, but was {Describe(either)}.");`

Every other assertion (`IsSome` :51, `IsRight` :86, `IsLeft` :110) uses
`throw new AssertionException(...)`. The split is exactly the one the commit
subject names: the assertions that return a value throw, the two `void` ones call
`Fail`.

On the tests, `grep -n "Multiple" Crypter.Test/Common_Tests/Assert_Tests.cs`
returns three fixtures and no more:

- `Is_Some_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` (:180)
- `Is_Right_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` (:194)
- `Is_Left_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` (:208)

All three cover value-returning assertions, and each asserts the *opposite* of
what `Fail` does — that the statement after the assertion never runs
(`Assert.That(returned, Is.False)`). Neither `IsNone` nor `IsNeither` has a
counterpart. Their only rejection tests, `Is_None_Fails_For_Some` (:92) and
`Is_Neither_Fails_For_Right` (:173), each call the assertion bare inside
`Assert.Throws<AssertionException>`, with no `Multiple` scope — the state in
which `Fail` and `throw` are the same thing.

**Mutation run.** Baseline for the fixture:
`dotnet test Crypter.Test --filter FullyQualifiedName~Assert_Tests` →
`Passed! - Failed: 0, Passed: 37`. Rewriting both `Fail` calls (lines 75 and 134)
as `throw new AssertionException(...)` and rerunning gives
`Passed! - Failed: 0, Passed: 37` — byte-identical result. The suite does not
detect a change that removes the entire behavioural difference the `Fail`/`throw`
split exists to create. The claim's reported numbers reproduce exactly.

(Restore needed `-p:NuGetAudit=false`; the project's restore fails on a
pre-existing `NU1903` warning-as-error for `SSH.NET` 2025.1.0, unrelated to this
diff. Both runs used the same flag, so the comparison is sound.)

**Probe.** With `Assert.cs` restored to HEAD, a temporary fixture ran each
assertion inside a `Multiple` scope:

```
PROBE IsNone:    continued=True, exceptionType=MultipleAssertException
PROBE IsNeither: continued=True, exceptionType=MultipleAssertException
```

Both assertions returned normally, the following statement executed, and the
scope ended in `MultipleAssertException` — not `AssertionException`. The
behaviour is real, is distinct from an explicit throw, and is reached by no test.
The probe file was removed; `git status --porcelain` is empty.

</details>

Crypter.Test/Assert.cs
Jack-Edwards: **maintainability-01** · maintainability lens · verifier: holds

#### What is broken

`Crypter.Test/Assert.cs:61` (`IsSome<T>(Maybe<T>, T)`), `:96` (`IsRight<TLeft, TRight>(Either, TRight)`)
and `:120` (`IsLeft<TLeft, TRight>(Either, TLeft)`) check the value with `That(...)` and then
`return value` unconditionally. Inside `Assert.Multiple`, `That` records the mismatch and execution
continues, so the caller receives the rejected value and keeps using it — verified above as
`returned=True observed=Bilbo`. The single-argument overloads on the same type abort instead, as the
branch's second commit and the three `Is_*_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` tests
intend.

Two things fix it, and both are needed for the rule to survive:

1. In each of the three expected-value overloads, compare explicitly and
   `throw new AssertionException(...)` on mismatch instead of calling `That`, matching how the
   single-argument overloads at lines 51, 86 and 110 already fail. The async wrappers need no change;
   they forward.
2. State the rule in the class `<remarks>` at lines 36-41 — an assertion on this type that returns a
   value throws rather than records, because a recorded failure inside `Assert.Multiple` would let the
   test continue with a value the assertion rejected. That sentence is the only thing that tells the
   next person adding an overload which half of the file to copy.

Adding expected-value cases to the three multiple-scope tests at lines 179-218 would pin it.

One correction to the claim's framing: it says a value-returning assertion built on `Fail` would hand
the test `default!`. For the expected-value overloads the returned value is not `default!` — the
preceding single-argument call already unwrapped a real value, so what leaks is the actual wrong value
(`"Bilbo"`), not a null. That makes the symptom quieter than the claim suggests, not louder: rather
than a `NullReferenceException` at the point of use, downstream code silently proceeds on a value the
assertion reported as wrong. The defect and the direction are unchanged.

<details><summary>The finding and evidence</summary>

> `Assert.cs` is built on an invariant it never states: an assertion that returns a value must throw
> rather than record. `IsSome(Maybe<T>)`, `IsRight(Either)` and `IsLeft(Either)` each
> `throw new AssertionException(...)` directly, while the void `IsNone` and `IsNeither` call `Fail(...)`.
> That split is load-bearing rather than incidental. Inside `Assert.Multiple`, NUnit's `Fail` records the
> failure and lets execution continue, so a value-returning assertion built on `Fail` would hand the test
> `default!` after reporting a failure. The branch's second commit ("Throw from the monad assertions that
> return a value") and the three tests named `Is_*_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` exist
> for exactly this reason.
>
> The rule appears nowhere in `Assert.cs`. The class `<remarks>` says only that the monad assertions
> return the matched value so a test can assert state and bind in one statement — nothing about how they
> must fail. And three of the six value-returning overloads already break it. `IsSome(maybe,
> expectedValue)` (the line this claim is anchored to), `IsRight(either, expectedValue)` and
> `IsLeft(either, expectedValue)` all check the value with `That(value, Is.EqualTo(expectedValue))` and
> then `return value` unconditionally. `That` records rather than throws inside a multiple scope, so these
> three do precisely what the invariant exists to prevent.
>
> Concrete failure. With `maybe` holding `Some("Bilbo")`:
>
>     Assert.Multiple(() =>
>     {
>         string name = Assert.IsSome(maybe, "Frodo");
>         Use(name);
>     });
>
> The mismatch is recorded, the method returns `"Bilbo"`, and `Use` runs on the value the assertion just
> rejected. […] The fixture does not catch this: all three multiple-scope tests use single-argument
> overloads, so the expected-value path is untested against `Assert.Multiple`.
>
> Direction: state the rule in the class `<remarks>` […] then make the expected-value overloads obey it
> — compare and `throw new AssertionException(...)` on mismatch instead of delegating the comparison to
> `That`. Extending the three multiple-scope tests to cover the expected-value overloads would keep it
> from drifting back.

Both files the claim names are added whole by this branch — `git diff origin/stable...HEAD --stat`
reports `Crypter.Test/Assert.cs | 223 ++++` and `Crypter.Test/Common_Tests/Assert_Tests.cs | 332 ++++`,
so this is code the diff introduced, not pre-existing code.

**The code is as described.** `Crypter.Test/Assert.cs`:

- The single-argument value-returning overloads throw directly: `IsSome` at line 51, `IsRight` at
  line 86, `IsLeft` at line 110 each `throw new AssertionException(...)`.
- The void assertions record: `IsNone` calls `Fail(...)` at line 75, `IsNeither` at line 134.
- The three expected-value overloads delegate the comparison to `That` and then return
  unconditionally: `IsSome` at lines 61-66 (`That(value, Is.EqualTo(expectedValue)); return value;`),
  `IsRight` at lines 96-101, `IsLeft` at lines 120-125. The async wrappers at lines 150, 175 and 192
  forward straight to them, so they inherit the same behaviour.
- The class `<remarks>` at lines 36-41 says only that the monad assertions return the matched value so
  a test can assert state and bind in one statement. Nothing about how they must fail.

**The failure reproduces.** I built `Crypter.Test` (`dotnet build Crypter.Test/Crypter.Test.csproj
-p:NuGetAudit=false` — the `-p:NuGetAudit=false` only suppresses an unrelated pre-existing NU1903
warning-as-error on `SSH.NET`) and ran a temporary probe fixture against the real `Assert` type, then
deleted it; the worktree is back to a clean `git status`. NUnit is 4.6.1 per
`Crypter.Test/Crypter.Test.csproj:18`.

Probe A — the expected-value overload, exactly the claim's scenario:

```csharp
Maybe<string> some = "Bilbo";
Assert.Multiple(() =>
{
    string name = Assert.IsSome(some, "Frodo");
    returned = true;
    observed = name;
});
```

Result: `returned=True observed=Bilbo`. The lambda continued past the failed assertion and was handed
the value the assertion had just rejected.

Probe B — the single-argument sibling, same scope shape, `Maybe<string>.None`:

```csharp
Assert.Multiple(() => { _ = Assert.IsSome(none); returned = true; });
```

Result: `returned=False`. The lambda aborted at the assertion.

So the two overloads of the same name behave oppositely inside `Assert.Multiple`, and the direction is
the one the claim states: the overload that checks *more* is the one that lets the test carry on with a
rejected value.

**The fixture does not cover it.** `Crypter.Test/Common_Tests/Assert_Tests.cs:179-218` holds all three
`Is_*_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` tests, and each one calls a single-argument
overload — `Assert.IsSome(None)` (line 184), `Assert.IsRight(Left)` (line 198), `Assert.IsLeft(Right)`
(line 212). No test puts an expected-value overload inside a multiple scope. The expected-value
overloads are tested only outside one, where `That` does throw
(`Is_*_With_An_Expected_Value_Fails_For_A_Different_Value`, e.g. line 162), which is why the gap is
invisible from a green run.

</details>

Crypter.Test/Assert.cs
```diff
@@ -0,0 +1,223 @@
...
+        return maybe.SomeOrDefault()!;
+    }
+
+    /// <summary>
+    /// Assert that the maybe is in the Some state and holds <paramref name="expectedValue"/>, then
+    /// return the value it holds.
+    /// </summary>
+    internal static T IsSome<T>(Maybe<T> maybe, T expectedValue)
+    {
+        T value = IsSome(maybe);
```
Jack-Edwards: **mutation-01** · mutation lens · verifier: holds

#### Where it holds

The three expected-value overloads at `Crypter.Test/Assert.cs:63`, `:98` and `:122` have no test
holding their state check. All three mutations survive the full fixture. Because these helpers
exist so a test can assert a monad's state and bind its value in one statement, a regression here
is silent: the helper returns a plausible value and the test that depended on it goes green while
asserting nothing about state.

The claim's suggested fix was verified as well. Three rejection tests —
`Is_Some_With_An_Expected_Value_Fails_For_None` expecting `"Expected Some, but was None."`,
`Is_Right_With_An_Expected_Value_Fails_For_Left` expecting `"Expected Right, but was Left(42)."`,
and `Is_Left_With_An_Expected_Value_Fails_For_Right` expecting `"Expected Left, but was Right(Frodo)."`
— fail 3/3 against the mutated code and pass 3/3 against the code as it stands. Asserting the
state message is what does the work: a dropped state check produces the equality-failure message
instead, so the equality assertion on the following line cannot satisfy the test on its own.

<details><summary>The finding and evidence</summary>

> The three expected-value overloads never have their state check held by a test, so an
> implementation that reads the payload without first checking the state passes the whole suite.
>
> `IsSome<T>(Maybe<T>, T)`, `IsRight<TLeft, TRight>(Either, TRight)` and `IsLeft<TLeft, TRight>(Either, TLeft)`
> each begin by delegating to the single-argument overload, which is what rejects the wrong state.
> [...] Mutation applied, one per overload, each the smallest edit that still compiles — replace the
> delegation with a direct payload read, dropping the state assertion.
>
> The failure this lets through: under the mutation `Assert.IsSome(Maybe<int>.None, 0)` passes.

The code is as described. `Crypter.Test/Assert.cs:61-66`, `:96-101` and `:120-125` are the three
expected-value overloads, and each one's state check is inherited entirely from the delegation on
its first line:

```csharp
internal static T IsSome<T>(Maybe<T> maybe, T expectedValue)
{
    T value = IsSome(maybe);          // Assert.cs:63 — the only state check
    That(value, Is.EqualTo(expectedValue));
    return value;
}
```

Both files are new in this diff (`git diff origin/stable...HEAD --stat` is `Crypter.Test/Assert.cs`
+223 and `Crypter.Test/Common_Tests/Assert_Tests.cs` +332, no other files), so the overloads and
their tests are the diff's own work.

The call-site claim checks out. Grepping the repository for `Assert.IsSome` / `Assert.IsRight` /
`Assert.IsLeft` and the `*Async` forms returns only `Crypter.Test/Assert.cs` and
`Crypter.Test/Common_Tests/Assert_Tests.cs`; nothing outside the fixture reaches these lines.

Within the fixture, the six tests that reach the expected-value overloads directly are
`Assert_Tests.cs:72-83` (`Is_Some_With_An_Expected_Value_Returns_The_Value` /
`..._Fails_For_A_Different_Value`), `:119-130` for `Is_Right_`, and `:153-164` for `Is_Left_`. The
async pairs at `:236-247`, `:278-289` and `:306-317` delegate to the same sync overloads, so they
are the same coverage. Every one of them passes a monad already in the expected state — `Some`,
`Right`, `Left` — and the three mismatch tests assert only `Assert.Throws<AssertionException>(...)`
with no message assertion, unlike the single-argument rejection tests at `:66-70`, `:105-110` and
`:139-144`, which do pin the message. Nothing in the file calls a two-argument overload with the
wrong state.

Run against the code in the worktree, with the three mutations applied exactly as the claim
specifies (`maybe.SomeOrDefault()!`, `either.RightOrDefault(default!)!`,
`either.LeftOrDefault(default!)!`):

- unmutated: `dotnet test Crypter.Test --filter FullyQualifiedName~Assert_Tests` → 37/37 passed
- all three mutations applied: → 37/37 passed

(`-p:NuGetAudit=false` was needed on both runs; restore fails on a pre-existing `SSH.NET` advisory
promoted to an error, unrelated to this diff.)

The concrete failure was reproduced, not just reasoned about. With the mutations in place, a
temporary fixture asserting the claim's three scenarios passed 3/3:

```csharp
int value = Assert.IsSome(Maybe<int>.None, 0);                        // passes, value == 0
int value = Assert.IsRight(Either<string, int>.FromLeft("boom"), 0);  // passes, value == 0
int value = Assert.IsLeft(Either<int, string>.FromRight("boom"), 0);  // passes, value == 0
```

`SomeOrDefault()` / `RightOrDefault(default!)` / `LeftOrDefault(default!)` yield `default(int)`,
which equals the expected `0`, so the equality assertion that follows is satisfied and the helper
returns cleanly from a monad in the state it exists to reject. For a reference-type payload the
mutation happens to be caught by the equality check against `null`, which is why the existing
`string`-valued mismatch tests do not notice it — the gap is silent precisely where the payload
type has a non-null default.

</details>

Crypter.Test/Assert.cs
```diff
@@ -0,0 +1,223 @@
...
+/// value in one statement.
+/// </remarks>
+internal abstract class Assert : NUnit.Framework.Assert
+{
+    /// <summary>
+    /// Assert that the maybe is in the Some state and return the value it holds.
+    /// </summary>
+    internal static T IsSome<T>(Maybe<T> maybe)
+    {
+        if (maybe.IsNone)
```
Jack-Edwards: **mutation-02** · mutation lens · verifier: holds

#### Where it holds

`Assert.IsSome` and `Assert.IsRight` have no test that distinguishes checking the monad's state
from checking its payload for null, because the fixture instantiates `T` and `TRight` as `string`
and nothing else. A slip that writes the check against the payload — the most likely slip here,
since both methods already call `SomeOrDefault()` / `RightOrDefault(default!)` on the next line —
ships green.

What that costs once the helpers are used: with a value-typed payload, a mutated `IsSome` returns
`default(T)` from a `None` without failing, so a test written as
`int count = Assert.IsSome(await GetCountAsync());` would pass against a `None` and then go on to
assert against a fabricated `0`. The same holds for `IsRight` over a left-state
`Either<string, int>`. `Maybe<T>` and `Either<TLeft, TRight>` are this repository's standard way of
carrying expected failures, so value-typed payloads — counts, ids, enums, booleans — are squarely
what these assertions will be pointed at, and that is the one payload shape the fixture never
covers. `IsLeft` is already covered for this by accident of `TLeft` being `int`; `IsSome` and
`IsRight` are the asymmetry.

Closing it: exercise the Some and Right rejection paths against a value-typed payload alongside the
existing reference-typed ones — a `Maybe<int>` and an `Either<string, int>` member next to the
existing ones, then assert that `IsSome` rejects `Maybe<int>.None` and that `IsRight` rejects a
left-state `Either<string, int>`. Step 4 above confirms those two assertions kill the mutation.

<details><summary>The finding and evidence</summary>

> `IsSome` and `IsRight` are only ever exercised with a reference-type payload, so a state check
> that tests the payload for null instead of testing the monad's state passes the whole suite.
>
> [...] Mutation applied, replacing each state check with a null check on the payload - the smallest
> edit that still compiles, and a plausible implementation slip given both helpers already reach for
> `SomeOrDefault` / `RightOrDefault` on the line below:
>
> - `if (maybe.IsNone)` becomes `if (maybe.SomeOrDefault() is null)`
> - `if (!either.IsRight)` becomes `if (either.RightOrDefault(default!) is null)`
>
> `dotnet test Crypter.Test --filter FullyQualifiedName~Assert_Tests` passes 37/37. [...] The
> identical mutation on `IsLeft` [...] fails 4 tests [...]
>
> The failure this lets through: under the mutation `Assert.IsSome(Maybe<int>.None)` returns `0`
> without failing [...] `Assert.IsRight(Either<string, int>.FromLeft("boom"))` returns `0` the same
> way.

Both files are new in this diff — `git diff origin/stable...HEAD --stat` shows
`Crypter.Test/Assert.cs` (+223) and `Crypter.Test/Common_Tests/Assert_Tests.cs` (+332) as the only
changes, so the claim is about code the change introduces.

The state checks are where the claim says they are: `Crypter.Test/Assert.cs:49`
(`if (maybe.IsNone)` in `IsSome<T>`) and `Crypter.Test/Assert.cs:84` (`if (!either.IsRight)` in
`IsRight<TLeft, TRight>`). Each returns the payload on the line below via `SomeOrDefault()` /
`RightOrDefault(default!)`.

The payload types in `Assert_Tests` are pinned once, at `Assert_Tests.cs:51-56`:

```csharp
private static Maybe<string> Some => Maybe<string>.From(SomeValue);
private static Maybe<string> None => Maybe<string>.None;
private static Either<int, string> Right => Either<int, string>.FromRight(SomeValue);
private static Either<int, string> Left => Either<int, string>.FromLeft(LeftValue);
private static Either<int, string> Neither => Either<int, string>.Neither;
```

Every `Maybe` test in the fixture goes through `Some` / `None`, so `T` is always `string`; every
`Either` test goes through `Right` / `Left` / `Neither`, so `TRight` is always `string` and `TLeft`
is always `int`. No test in the fixture instantiates either monad with a value-typed `T` or
`TRight`. I read all 37 tests to confirm this — there is no exception.

I ran the experiment in the worktree (`-p:NuGetAudit=false` is needed to restore at all; the
`SSH.NET` NU1903 advisory is escalated to an error and is unrelated to this diff):

1. **Baseline**, unmodified: `Passed! - Failed: 0, Passed: 37`.
2. **Mutation**, lines 49 and 84 replaced exactly as the claim specifies: `Passed! - Failed: 0,
   Passed: 37`. The mutation survives the entire suite.
3. **Control**, line 108 `if (!either.IsLeft)` replaced with
   `if (either.LeftOrDefault(default!) is null)`: `Failed! - Failed: 4, Passed: 33`, failing
   `Is_Left_Fails_For_Right`, `Is_Left_Fails_For_Neither`, `Is_Left_Async_Fails_For_Right` and
   `Is_Left_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` — the four the claim names, and only
   those. The sole difference is that `TLeft` is `int`, whose `default` is not null, so the
   reference-type payload is what lets the mutation through rather than anything else about the
   tests.
4. **The consequence**, as a temporary fixture against the mutated helpers:
   `Assert.IsSome(Maybe<int>.None)` and `Assert.IsRight(Either<string, int>.FromLeft("boom"))` both
   returned `0` and passed. Against the unmutated helpers the same two probes fail with
   `Expected Some, but was None.` and `Expected Right, but was Left(boom).`, which confirms a
   value-typed case is a genuine kill for the mutation and not merely a test that happens to fail.

The temporary fixture was removed and `Assert.cs` restored; `git status --porcelain` is empty.

</details>

Crypter.Test/Assert.cs
```diff
@@ -0,0 +1,223 @@
...
+/// so the inherited assertions are reached through the same <c>Assert</c> name they always were.
+/// The monad assertions return the matched value, which lets a test assert the state and bind the
+/// value in one statement.
+/// </remarks>
+internal abstract class Assert : NUnit.Framework.Assert
+{
+    /// <summary>
+    /// Assert that the maybe is in the Some state and return the value it holds.
+    /// </summary>
+    internal static T IsSome<T>(Maybe<T> maybe)
```
Jack-Edwards: **naming-01** · naming lens · verifier: holds

#### The failure, as it stands

`Crypter.Test/Assert.cs:47`. A test author writes the shape the rest of the type invites:

```csharp
Assert.Multiple(() =>
{
    Assert.IsSome(maybeProfile);
    Assert.That(user.Username, Is.EqualTo(expectedUsername));
    Assert.That(user.EmailAddress, Is.EqualTo(expectedEmail));
});
```

When `maybeProfile` is None, line 51 throws, the multiple scope aborts, and the two `Assert.That`
calls never run — so the report names the maybe and hides a wrong username or email address
underneath it. Swap `Assert.IsSome` for the sibling `Assert.IsNone` and the block collects all
three, because `IsNone` goes through `Fail`. Both are `Assert.Is*` calls on a monad; nothing at
either call site says which one you get.

The throwing behaviour is correct and should stay — there is no value to hand back once the state is
wrong. The fix is the name. Rename the three value-returning assertions and their six async
overloads so the name carries the payload and the abort — `SomeValueOf` / `RightValueOf` /
`LeftValueOf`, or an `Unwrap*` form — and leave `IsNone` and `IsNeither`, which genuinely are void
state checks, on the `Is*` shape. Once the name announces the return, the per-method doc sentence at
lines 45, 59, 80, 94, 104 and 118 and the `<remarks>` sentence at 39-40 stop being load-bearing.

<details><summary>Evidence</summary>

The diff (`git diff origin/stable...HEAD`) adds exactly two files: `Crypter.Test/Assert.cs` (223
lines, new) and `Crypter.Test/Common_Tests/Assert_Tests.cs` (332 lines, new). The snippet
`internal static T IsSome<T>(Maybe<T> maybe)` is at `Crypter.Test/Assert.cs:47`.

**The asymmetry is in the file and is deliberate.** `Crypter.Test/Assert.cs` declares six `Is*`
members on one type, split down the middle by failure semantics:

- Value-returning, abort on failure: `IsSome` (47), `IsRight` (82), `IsLeft` (106) — each
  `throw new AssertionException(...)` (51, 86, 110) and return the payload (54, 89, 113).
- Void, record on failure: `IsNone` (71) and `IsNeither` (130) — each call `Fail` (75, 134).

Nothing in the six names distinguishes the two groups. The commit message on `a8d331ef` states the
split as intent: "Inside an Assert.Multiple scope, Fail records the failure and returns rather than
throwing [...] There is no value to return once the state is wrong, so these now throw. IsNone and
IsNeither still call Fail, so a multiple scope collects them."

**The claimed consequence reproduces.** I built a standalone NUnit 4.6.1 (`Crypter.Test.csproj:18`)
probe mirroring the two shapes from `Assert.cs` — a helper that throws `AssertionException`, and one
that calls `Assert.Fail` — each inside `Assert.Multiple`:

- Throwing shape: `0` statements after the helper ran; the scope aborted and a bare
  `AssertionException` escaped. Only the monad failure is reported.
- Recording shape: `1` statement after the helper ran, the scope continued to a further failing
  `Assert.That`, and a `MultipleAssertException` reported **both** failures.

That is the same behaviour the branch pins for itself in
`Assert_Tests.Is_Some_Inside_A_Multiple_Scope_Does_Not_Return_A_Value`
(`Crypter.Test/Common_Tests/Assert_Tests.cs:180-191`), which asserts the statement after
`Assert.IsSome(None)` never executes, with the same shape for `IsRight` (194) and `IsLeft` (208).

**The token is taken.** Both cited sites check out: `Crypter.Core/Features/Transfer/Common.cs:72` is
`if (recipientUsername.IsSome && recipientId.IsNone)` and
`Crypter.Core/Services/TransferOwnershipService.cs:36` is `return recipientId.IsSome ? ... : ...`.
`.IsSome`/`.IsNone` appear 35 times outside the test project, uniformly as bool state predicates.

**The docs are carrying the name's weight.** All six value-returning methods carry a doc sentence
whose only content is the return ("...and return the value it holds", lines 45, 58-59, 80, 93-94,
104, 117-118, and the six async ones), and the type `<remarks>` needs a further sentence for it
("The monad assertions return the matched value...", 39-40).

</details>

Crypter.Test/Common_Tests/Assert_Tests.cs
```diff
@@ -0,0 +1,332 @@
...
+ * code without disclosing the source code of your own applications.
+ *
+ * Contact the current copyright holder to discuss commercial license options.
+ */
+
+using System.Threading.Tasks;
+using EasyMonads;
+using NUnit.Framework;
+
+namespace Crypter.Test.Common_Tests;
```
Jack-Edwards: **placement-01** · placement lens · verifier: holds

#### Where it holds

`Crypter.Test/Common_Tests/Assert_Tests.cs:31` places a test for `Crypter.Test.Assert` inside the
directory that mirrors the `Crypter.Common` project, under `namespace Crypter.Test.Common_Tests`.
`Crypter.Common` has no `Assert` type, and the file's subject — the test project's own NUnit-shadowing
assertion harness declared at `Crypter.Test/Assert.cs:42` — belongs to no project folder `Common_Tests`
mirrors.

The concrete cost: `Common_Tests/` is the directory a contributor reads to find out what of
`Crypter.Common` is covered, and its two existing subfolders both answer that question exactly. A
third entry that answers a different question breaks the only signal the folder carries. `Common_Tests`
is also incomplete — `Crypter.Common/Contracts`, `Mappers`, `Services`, `Enums`, `Attributes` and
`Exceptions` have no mirror subfolder yet — so whoever fills those in next has to decide what to do
with a file whose mirror counterpart does not and cannot exist.

The fix, and the repo's own precedent for it: move `Assert_Tests.cs` to a sibling top-level folder
under `Crypter.Test` alongside `Container_Tests/`, and set the namespace to match the new folder, the
way `Container_Tests/PostgresContainer_Tests.cs` does for the container harness. `Crypter.Test.Assert`
stays visible from the new namespace either way, since it is declared directly in `Crypter.Test` and
every `Crypter.Test.*` namespace is nested inside it.

<details><summary>Evidence</summary>

The diff adds exactly two files, so both the subject and its placement are this change's to own:

```
$ git diff --stat origin/stable...HEAD
 Crypter.Test/Assert.cs                    | 223 ++++++++++++++++++++
 Crypter.Test/Common_Tests/Assert_Tests.cs | 332 ++++++++++++++++++++++++++++++
```

**The subject is test-project code, not `Crypter.Common` code.** `Crypter.Test/Assert.cs:31` declares
`namespace Crypter.Test;` and `Crypter.Test/Assert.cs:42` declares
`internal abstract class Assert : NUnit.Framework.Assert`. The test file sits at
`Crypter.Test/Common_Tests/Assert_Tests.cs` with `namespace Crypter.Test.Common_Tests;` on line 31.

**The mirror is exact, as claimed.** `Crypter.Common/Infrastructure` contains `UrlSafeEncoder.cs` and
`VersionProvider.cs`; `Common_Tests/Infrastructure_Tests/` contains `UrlSafeEncoder_Tests.cs` and
`VersionProvider_Tests.cs` — a one-to-one correspondence. `Common_Tests/Primitive_Tests/` holds the
five primitive tests against `Crypter.Common/Primitives`. `Crypter.Core` has top-level `Features/`,
`Identity/` and `Services/`, matched by `Core_Tests/Features_Tests/`, `Core_Tests/Identity_Tests/`
and `Core_Tests/Services_Tests/`.

**The subject-mismatch predicate is true across the whole project.** `find Crypter.Test -name '*_Tests.cs'`
returns 63 files. Every one of the other 62 has its subject in the project or area its containing
folder is named for. `Assert_Tests.cs` is the only file under `Common_Tests/` whose subject is not in
`Crypter.Common`.

**The precedent is real.** `Crypter.Test/Container_Tests/PostgresContainer_Tests.cs:32` declares
`namespace Crypter.Test.Container_Tests;` and its body drives `AssemblySetup` and `ContainerService`
— root-level harness types from `Crypter.Test/AssemblySetup.cs` and `Crypter.Test/ContainerService.cs`,
in namespace `Crypter.Test`. That is test-project infrastructure given its own sibling top-level
`*_Tests` folder rather than a slot in another project's mirror, which is exactly the destination
shape the finding asks for.

**The supporting detail about the monads is correct.** `Crypter.Common/Crypter.Common.csproj:8` carries
`<PackageReference Include="EasyMonads" Version="2.5.0" />`, and `Crypter.Common/Monads` does not
exist (`ls` fails; the project's directories are `Attributes`, `Contracts`, `Enums`, `Exceptions`,
`Infrastructure`, `Mappers`, `Primitives`, `Services`). `Crypter.Test/Assert.cs` imports
`using EasyMonads;`. So the types the new assertions wrap are not `Crypter.Common` types either, and
`CLAUDE.md`'s reference to "`Maybe<T>` and `Either<TLeft, TRight>` from `Crypter.Common/Monads`" points
at a folder that is not there — which is the same wrong mental model the placement encodes.

</details>

Crypter.Test/Common_Tests/Assert_Tests.cs
Jack-Edwards: **restatement-01** · restatement lens · verifier: holds

#### Where it holds

The `<summary>` at `Crypter.Test/Common_Tests/Assert_Tests.cs:34` asserts a count — "exercised
twice" — that the fixture directly below it contradicts for three of its sixteen assertions. The
count was accurate when the summary was written and was invalidated by the second commit on this
same branch, which added five tests without revisiting the prose.

Nothing fails at run time; the cost is to the next person to touch the fixture. The summary is the
file's stated completeness rule, so a reader either caps a newly added assertion at two cases
because that is what the file says it does, or hunts for a missing pair for `IsRight` that was
never missing.

The durable fix is to drop the count and keep the content, which survives further additions:
"Every monad assertion is exercised against the state it expects and against a state it must
reject." Restating the current number instead just re-arms the same trap for the next commit that
adds a case.

<details><summary>Evidence</summary>

The diff is entirely additive against `origin/stable` — `git diff --stat origin/stable...HEAD`
reports 555 insertions and no deletions across `Crypter.Test/Assert.cs` (223 lines, new) and
`Crypter.Test/Common_Tests/Assert_Tests.cs` (332 lines, new). Both the summary sentence and every
test method it describes are introduced by this branch, so the finding is not describing untouched
code.

`grep -c "\[Test\]"` over the fixture returns **37**, not the 32 that sixteen assertions exercised
twice would produce. That gap of five is exactly the five methods added by a8d331ef
(`Is_Right_Fails_For_Neither`, `Is_Left_Fails_For_Neither` and the three
`*_Inside_A_Multiple_Scope_Does_Not_Return_A_Value` tests), and `git show --stat c025dfc0` confirms
the first commit landed the fixture at 32 tests with the summary already worded as it is now.

Reading the bodies at `Assert_Tests.cs:58-220` confirms the per-assertion counts:

- one-argument `Assert.IsSome` — three call sites: `Is_Some_Returns_The_Value` (line 61),
  `Is_Some_Fails_For_None` (line 68), `Is_Some_Inside_A_Multiple_Scope_Does_Not_Return_A_Value`
  (line 186).
- one-argument `Assert.IsRight` — four call sites: lines 101, 108, 115, 200.
- one-argument `Assert.IsLeft` — four call sites: lines 134, 141, 148, 214.

The two-argument overloads, `IsNone`, `IsNeither` and the whole async set are still exercised
exactly twice, which is what makes the sentence read as a rule rather than an obvious slip.

</details>

Jack-Edwards (commented):
This review was written by the Crypter pipeline's reviewer lenses, and each finding here was checked against the code at a8d331ef by a separate verifier before it was posted. Only findings the verifier ruled as holding appear below; each comment carries the verifier's ruling, with its evidence collapsed.

**Lenses that ran:** comments, correctness, coverage, directness, documentation, maintainability, mutation, naming, placement, restatement, security, testability.

**Lenses that raised nothing:** documentation, security, testability.

All 11 held findings are on lines in the diff. Several describe the same underlying point from different lenses — the throw-versus-`Fail` split inside `Assert.Multiple` (comments-01, correctness-01, maintainability-01, naming-01, coverage-01/02) and the "exercised twice" fixture summary (comments-03, restatement-01).

The lenses read the diff, not the discussion around it, so a finding that rests on an assumption about intent says so in its text.

PR #853, opened 2026-08-06
Authenticate the pipeline container by login or by token

Jack-Edwards:
Claude Code in the pipeline container authenticates from an interactive login held in the `crypter-pipeline-claude` volume. Renewing it when it lapses meant knowing to open a shell in the container and run `claude` there.

`pipeline.sh login` makes that a command. `CLAUDE_CODE_OAUTH_TOKEN` in `.devcontainer/.env` is an alternative for a machine that would rather configure the credential than open a browser; Compose reads it into the container's environment, where Claude Code uses it in place of the stored login.

Nothing is required of you after merging — an existing login keeps working. `up` warns when neither a token nor a login is configured, rather than refusing.


Jack-Edwards:
Closing unmerged. The development pipeline now lives in [Crypter-File-Transfer/Claude-Development-Pipeline](https://github.com/Crypter-File-Transfer/Claude-Development-Pipeline), and this branch's changes went across with it — `.devcontainer/.env.example`, `.devcontainer/docker-compose.yml`, `.devcontainer/pipeline.sh` and the pipeline document are byte-identical there.

`stable` no longer carries the pipeline, so there is nothing here left to merge into.

PR #854, opened 2026-08-13
Build Crypter.Web on Node 26

Jack-Edwards:
The web image installs Node 22, which leaves security support in April 2027. Node 26 becomes LTS in October 2026 and is supported until April 2029.

NodeSource serves 26.7.0 for the `dotnet/sdk:10.0` base the build stage uses, and pnpm 11.18.0 installs and runs `pnpm setup` under it. The image build in CI covers the rest.

### DIFF
--- a/Crypter.Web/Dockerfile
+++ b/Crypter.Web/Dockerfile
@@ -7,7 +7,7 @@ WORKDIR /source/
 SHELL ["/bin/bash", "-c"]
 
 RUN apt-get update -yq && apt-get upgrade -yq && apt-get install -yq curl git nano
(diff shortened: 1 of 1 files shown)


Jack-Edwards:
The three red checks are all the same pre-existing restore failure, not this change:

```
Crypter.Test.csproj : error NU1903: Warning As Error: Package 'SSH.NET' 2025.1.0 has a
known high severity vulnerability
```

`build-and-test` and both CodeQL `Analyze` jobs stop there, before anything is compiled. `stable` last ran this workflow on 2026-08-06 and the advisory has been published since, so it fails on unmodified `stable` too. It wants its own pull request.

`build-web`, which is the job that builds the Dockerfile this change edits, passed.

PR #859, opened 2026-09-21
Resolve integration test file storage from the repo root

Jack-Edwards:
`Crypter.Test` read its transfer storage location from `IntegrationTestingOnly:TransferStorageLocation`, whose Linux value was the GitHub Actions runner's path. Every other machine had to hand-edit the file, and the tests could not run anywhere the checkout is not at `/home/runner/work/Crypter`.

`AssemblySetup` now derives the location from `SettingsReader.GetRepoPath()`, the way `ContainerService` already derives the PostgreSQL init files. The setting and the per-OS branch that read it are gone, and `crypter_files/` is ignored.

A full `Crypter.Test` run is green: 206 passed, 0 failed.

One thing a reviewer would otherwise have to discover: Claude-Development-Pipeline clones its workspaces from this remote, so the change there that runs these tests inside the container needs this on `stable` first.


Jack-Edwards:
## Review

Read against the branch. Nothing to fix.

`GetRepoPath()` walks up from the assembly to the directory named after it and returns that directory's parent, which is the repo root — the same derivation `ContainerService.GetPostgresInitVolume()` already relies on, so the storage location and the PostgreSQL init files now come from one place rather than two.

Checked and clear:

- Nothing else reads `IntegrationTestingOnly:TransferStorageLocation`. A search of the repo returns the two hits this diff deletes.
- The removed `Microsoft.Extensions.Configuration` using has no other user in `AssemblySetup.cs`.
- The old Linux value sat above the CI checkout at `/home/runner/work/Crypter/`, and the new location is inside it, which is what makes the `crypter_files/` ignore entry necessary rather than tidy.
- `FileStorageLocation` is `string?` and `Path.Join` answers non-null, so the field is no longer nullable in practice; `ResetServerDataAsync`'s `Directory.Delete(FileStorageLocation!, true)` already swallows a missing directory.

### The `SSH.NET` advisory needs an issue

The description names it under what a reviewer would otherwise discover, so it goes when this merges. It is load-bearing for the pipeline: Claude-Development-Pipeline#125 tells its implementer to run `dotnet test Crypter.Test` unfiltered in the container, and `Crypter.Test.csproj:5` sets `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` unconditionally, so `NU1903` stops the restore. Every agent run meets it until the unused `TestContainers.Container.Database.PostgreSql` 1.5.4 goes.

File it and link it from the description.

### Merge order

Right as stated, and worth keeping in view: this to `stable` before Claude-Development-Pipeline#125, since workspaces are cloned from the remote.

Jack-Edwards:
## Correction: the `SSH.NET` advisory has lapsed

My comment above repeated this PR's `NU1903` note and asked for an issue. Having actually run it, there is nothing to file:

- `dotnet restore Crypter.Test` on a fresh shallow clone of `stable` with an empty package cache succeeds, and `-p:NuGetAuditMode=all` reports no advisory.
- `SSH.NET` resolves to 2026.0.0 rather than the 2025.1.0 recorded here.
- It comes from `Testcontainers` 4.15.0 — declared there in `project.assets.json` — not from the unused `TestContainers.Container.Database.PostgreSql`, so removing that package was never going to move it.

The description's second bullet is stale rather than wrong: it was true when written. Worth dropping from the description now, along with the `-p:NuGetAudit=false` the runs used, since neither is needed to reproduce them.

The unused package is still worth removing for its own sake, and that is #862.

The rest of the review stands: nothing to fix in the change itself, and the merge order holds — this to `stable` before Claude-Development-Pipeline#125.

